PR #16, opened 2012-07-09
Build minissdpd with autotools

crrodriguez:
The following request allows building minissdpd with autotools to ease distributors work.


miniupnp:
I still need to make it work for me ;)

miniupnp:
I never had the time to make it work properly

PR #41, opened 2013-07-11
Add PCP support

ghost:
- PCP MAP operation supported on all platforms (to enable run: ./genconfig.sh --pcp)
- optional PCP PEER operation implemented only on Linux/netfilter platform (./genconfig.sh --pcp-peer)


apoorv-2204:
hello

PR #99, opened 2015-01-21
Fix setsockopt() for IPV6_MULTICAST_IF

maksqwe:
32bit: sizeof(&ifindex) == sizeof(void *) == 4
64bit: sizeof(&ifindex) == sizeof(void *) == 8

As a result, it is "ifindex" oob acces on 64bit platforms.


miniupnp:
see 81dc08d2907efe23389d8af8c83aebca769f8278

PR #103, opened 2015-03-09
Fix get_portmappings_in_range() in non-expand case.

s1061123:
In get_portmappings_in_range(), array[] is not updated when realloc() is not called, 
hence get_portmappings_in_range() is always null.  This fix changes to fill array[].


s1061123:
Hi folks, 

It seems to be bug. Could you please take a look into it?

Issue #104, opened 2015-04-21
Redirect outside port X to port Y on 'this' IP address


sanderjo:
I want to redirect outside port X to port Y on 'this' IP address (with X not the same as Y), so command "-d" does not do the trick.

As far as I know, there is no direct command for that as you first have find out the RFC1918 address on the host you're on. I now use:

```
upnpc -a $(ip route get 8.8.8.8 | head -1 | awk '{ print $NF }') 22 22000 tcp
```
to redirect outside port 22000 to port 22 on the system issueing the command.

If the above assumption is correct, I suggest an extra option "-m"  (for: "me"):

```
./upnpc-static [options] -m port external_port protocol [duration]
```
With that implemented, the usage would be:
```
upnpc -m 22 22000 tcp
```

Implementing should be feasible as upnpc already knows it's own RFC1918 IP address

miniupnp:
there is already a -m option ! why not modify the -r option, or make a specific executable suitable for your use.

sanderjo:
Ah ... I missed the "-m" at the bottom. Sorry for that. How about "-b"?

I was able to create what I wanted:

Normal port redirect:
```
sander@flappie:~/git/miniupnp/miniupnpc$ ./upnpc-static -b 555 55555 tcp
upnpc : miniupnpc library test client. (c) 2005-2014 Thomas Bernard
Go to http://miniupnp.free.fr/ or http://miniupnp.tuxfamily.org/
for more information.
List of UPNP devices found on the network :
 desc: http://192.168.1.1:5000/rootDesc.xml
 st: urn:schemas-upnp-org:device:InternetGatewayDevice:1

Found valid IGD : http://192.168.1.1:5000/ctl/IPConn
Local LAN ip address : 192.168.1.113
ExternalIPAddress = 3.128.233.31
InternalIP:Port = 192.168.1.113:555
external 3.128.233.31:55555 TCP is redirected to internal 192.168.1.113:555 (duration=0)

```


Redirect with duration:
```
sander@flappie:~/git/miniupnp/miniupnpc$ ./upnpc-static -b 44 4444 tcp 100
upnpc : miniupnpc library test client. (c) 2005-2014 Thomas Bernard
Go to http://miniupnp.free.fr/ or http://miniupnp.tuxfamily.org/
for more information.
List of UPNP devices found on the network :
 desc: http://192.168.1.1:5000/rootDesc.xml
 st: urn:schemas-upnp-org:service:WANIPConnection:1

[...]

PR #106, opened 2015-04-21
Add option "-b" for redirect port X to port Y to 'this' system

sanderjo:
Add option "-b" for redirect port X to port Y to 'this' system (= system issuing the UPnP command).

See https://github.com/miniupnp/miniupnp/issues/104


miniupnp:
#108 is better :)

PR #108, opened 2015-04-24
add optional parameter <external_port> to option -r

sanderjo:
As discussed https://github.com/miniupnp/miniupnp/issues/104#issuecomment-95973663 and https://github.com/miniupnp/miniupnp/issues/104#issuecomment-95973663

### DIFF
--- a/miniupnpc/upnpc.c
+++ b/miniupnpc/upnpc.c
@@ -41,6 +41,18 @@ const char * protofix(const char * proto)
 	return 0;
 }
 
+/* is_int() checks if parameter is an integer or not
+ * 1 for integer
+ * 0 for not an integer */
+int is_int(char const* s) {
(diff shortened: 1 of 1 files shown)


sanderjo:
Interesting catch. Although I don't think that problem can occur within the context of upnpc.c (as commandargv uses spaces as seperator), I'll change it. Hopefully it will arrive in the same pull request.

But maybe the is_int() is not needed at all? Instead of
```
if(!is_int(commandargv[i+1])){
```
maybe easier to use
```
if(protofix(commandargv[i+1])){
```
It does not test if the 2nd parameter is a number, but it tests if the 2nd parameter is 'tcp' or 'udp'. Same same, but different. ;-)

That would eliminate the need for is_int() at all, so cleaner code.

Feedback welcome.

PR #129, opened 2015-06-15
Make Tomato's OS_VERSION shorter to stop SendSSDPNotify log messages

edrikk:
As discussed in this thread:
http://www.linksysinfo.org/index.php?threads/tomato-shibbys-releases.33858/page-53#post-262636

It appears that the Tomato OS_VERSION has gotten long enough, that it's causing SSDP_PACKET_MAX_LEN being larger than 512 by a few characters.

This change reduces unnecessary "verbiage" at the end of the Tomato OS_VERSION string.

In short, what was previously (as example):
     Tomato 1.28.0000 MIPSR2-130 K26AC USB AIO-64K
 will become
     Tomato MIPSR2-130 K26AC

### DIFF
--- a/miniupnpd/genconfig.sh
+++ b/miniupnpd/genconfig.sh
@@ -73,7 +73,8 @@ fi
 # Tomato USB special case
 if [ -f ../shared/tomato_version ]; then
 	OS_NAME=Tomato
-	OS_VERSION="Tomato $(cat ../shared/tomato_version)"
+	TOMATO_VER=`cat ../shared/tomato_version | cut -d' ' -f2,3`
+	OS_VERSION="Tomato $TOMATO_VER"
 fi
 
 ${RM} ${CONFIGFILE}


koitsu:
I should note that this issue is, in my opinion, still a problem.  Shortening OS_VERSION ("SERVER" header in SSDP NOTIFY solicitations) does not solve the problem long-term.  Analysis with evidence is provided here: http://www.linksysinfo.org/index.php?threads/tomato-shibbys-releases.33858/page-58#post-267640

I would strongly suggest SSDP_PACKET_MAX_LEN be set to 1024 universally (i.e. remove the #ifdef ENABLE_HTTPS and just #define SSDP_PACKET_MAX_LEN 1024 regardless).  If a new pull request is needed for this, just ask and I can submit one.

edrikk:
I agree that increasing the size is another solution, but I wouldn't call it a better solution;  In my opinion then the "next" implementation will just go over that boundary...  I believe it's the responsibility of the implementer to respect the solution's boundaries.

A better solution (if the standard allows for it) would be to COMPLETELY REMOVE the upper bound if it's no longer relevant today (or mute the logging if miniupnpd wants to be strict against the protocol, but it's known not to cause issues).

PS.  These changes have not yet been merged into Tomato by Shibby as yet, which is why the users are still seeing the log messages in Tomato as of late October 2015.

edrikk:
(To be clearer:  by "completely remove" I mean dynamic allocation/reallocation off buffer size vs hardcoded size).

miniupnp:
anyway there is an "hard" limit which is maximum UDP payload size, slighly below network MTU
going anywhere above 1400bytes is risky

PR #134, opened 2015-07-08
Moving to setuptools to enable wheel

vsapronov:
This move to setuptools allows to build (relatively) new binary wheel distribution and keep current source distribution (https://pypi.python.org/pypi/miniupnpc/1.9) as well.

The miniupnpc is done in native code. In many cases you just want to get compiled working binaries compatible with the platform without recompiling them. This is critical for ARM processors, since compile time there is much longer.

The choice of setuptools and bdist_wheel is inline with latest python packaging recommendations: https://packaging.python.org/en/latest/current.html.

### DIFF
--- a/miniupnpc/setup.py
+++ b/miniupnpc/setup.py
@@ -6,7 +6,7 @@
 # python script to build the miniupnpc module under unix
 #
 # replace libminiupnpc.a by libminiupnpc.so for shared library usage
-from distutils.core import setup, Extension
+from setuptools import setup, Extension
 from distutils import sysconfig
 sysconfig.get_config_vars()["OPT"] = ''
 sysconfig.get_config_vars()["CFLAGS"] = ''


miniupnp:
just for my information, what will it change for existing user ?
For which version of python does it work, etc ?

vsapronov:
For the existing "user" (clearly developer) process won't change at all. With setuptools you can keep building and deploying (to PyPi) source distributions with `python setup.py sdist` as you have done before. This tar.gz source distributions will be exactly the same as the one created by distutils and they are installable with `pip install miniupnpc`. At that point pip will download sources, compile C-code and put into appropriate place in site-packages. Everything as usual so far.

New thing added is a possibility to build `python setup.py bdist-wheel` which will produce binary wheel package. Then the one can create one for each platform: python27, python33, linux, windows, x86, x64, ARM or at least for some of them (most popular). They will be accessible altogether with source tar.gz package. Then pip wheel can install from such binary package (if platform matches of course) without invoking compiler on the target machine.

Check psutil package to see example of such package: https://pypi.python.org/pypi?:action=display&name=psutil#downloads. Pay attention to .whl files altogether with tar.gz. You also might be interested to see this wheel migration status for top-360 python packages: http://pythonwheels.com/.

PR #142, opened 2015-07-21
Read USN from SSDP messages.

annulen:
USN is mandatory header so nothing should break.
Yes, this breaks ABI of miniupnpc, but I think it's worth it (descURL is not guaranteed to be constant, and I need unique service identifiers)
Note that I've changed listdevices to print USN instead of ST - may be not what you want in this tool.


annulen:
Just realized that it's sufficient to put usn below scope_id to preserve ABI compatibility, because size of UPNPDev is not known in advance anyway, and nobody sane will access buffer directly

PR #146, opened 2015-08-04
Moved SSDP discovery code into minissdpc.c

annulen:
Currently this patchset breaks non-POSIZ platforms so rigth now it's more like proof of concept than actual pull request.

This refactoring allows me to use minissdpc,receivedata, and upnpdev as self-contained SSDP client library which can use either minissdpd or in-process SSDP disovery, without including any code related to HTTP, XML, SOAP and IGD-specific functions. Also I use persistent connection with minissdpd (which otheriwse would require passing of file descriptor instead of minissdpdsock to upnpDiscoverDevices)

If you don't object to such refactoring I'll try to fix non-POSIX platforms and other build systems


annulen:
ping

PR #157, opened 2015-10-25
String handling fixes

laanwj:
I found these bugs during a security audit of miniupnpc in the context of TALOS-2015-0035 in Bitcoin Core, and tried to fix them.

Neither of these are remote vulnerabilities, as they are triggered by input from the application not from the network. However they lead to exposure of memory contents or stack overflows when used as part of a multi-pronged exploit.

- Check return value of snprintf: snprintf on Linux returns the number of characters that would have been written if there was enough space in the buffer. Return an error condition if space was not sufficient.

- Account exactly for bytes when building buffer in simpleUPnPcommand2. The margin of 100 is not guaranteed to always be enough. When long parameters are passed in, it was possible to overflow the buffer.

### DIFF
--- a/miniupnpc/minisoap.c
+++ b/miniupnpc/minisoap.c
@@ -105,6 +105,8 @@ int soapPostSubmit(int fd,
 					   "Pragma: no-cache\r\n"
 					   "\r\n",
 					   url, httpversion, host, portstr, bodysize, action);
+	if ((unsigned int)headerssize >= sizeof(headerbuf))
+		return 0;
 #ifdef DEBUG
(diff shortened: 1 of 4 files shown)


laanwj:
Can't explain the appveyor CI error, looks like a build system issue, but this change does not touch the build system:
```
sh: C:\Program: No such file or directory
Makefile.mingw:49: recipe for target 'miniupnpc.lib' failed
mingw32-make: *** [miniupnpc.lib] Error 127
Command exited with code 2
```

miniupnp:
I fixed the AppVeyor script

Issue #160, opened 2015-11-13
get_lan_for_peer broken for IPv6 on BSD?


razzfazz:
The pfSense folks are seeing an issue where miniupnpd doesn't accept IPv6 requests; see [here](https://redmine.pfsense.org/issues/4321#change-22410). Looking over the relevant code, I think the problem is that the BSD version of get_src_for_route_to will always return failure if the src and/or src_len parameters are NULL (as they are in get_lan_for_peer): The condition in the if statement in miniupnp/miniupnpd/bsd/getroute.c:87 will never evaluate to true in this case, and as a result "found" will never be set to 1, and the function will always return -1, causing get_lan_for_peer to fail.

miniupnp:
once you applied the patch, can you give me the output of
```
$ ./testgetroute 8.8.8.8
$ ./testgetroute 2a00:1450:4007:806::101f
```

razzfazz:
$ ./testgetroute 8.8.8.8
testgetroute: calling get_src_for_route_to(0x7fffffffea70, NULL, NULL, 0x7fffffffea44)
testgetroute: read l=192 seq=1 pid=86966
testgetroute: type=1 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute: type=2 sa_len=16 sa_family=2 [my WAN address]:0
testgetroute: type=4 sa_len=0 sa_family=0 unknown address family 0
testgetroute: get_src_for_route_to() returned 0
testgetroute: index=-1
testgetroute: calling get_src_for_route_to(0x7fffffffea70, 0x7fffffffea74, 0x7fffffffea48(4), 0x7fffffffea44)
testgetroute: read l=192 seq=1 pid=86966
testgetroute: type=1 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute: type=2 sa_len=16 sa_family=2 [my WAN address]:0
testgetroute: type=4 sa_len=0 sa_family=0 unknown address family 0
testgetroute: get_src_for_route_to() returned 0
testgetroute: src=[my WAN address]:0
testgetroute: index=-1

razzfazz:
$ ./testgetroute 2a00:1450:4007:806::101f
testgetroute: calling get_src_for_route_to(0x7fffffffea40, NULL, NULL, 0x7fffffffea34)
[hangs here]

Attaching a debugger shows that the hang is in the read() call inside get_src_for_route_to().

razzfazz:
[...]

PR #162, opened 2015-11-16
fix get_src_for_route_to() on BSD

razzfazz:
see issue #160


razzfazz:
Looks like this was already cherry-picked. Closing.

Issue #167, opened 2015-11-27
Since UPNP_VERSION_MAJOR was set to 2, Windows 10 BITS client generates event errors


RMerl:
I recently upgraded to the latest miniupnpd code, where UPNP_VERSION_MAJOR was increased from "1" to "2".  Since then, I get these a few times per day in Windows's event log:

```
BITS has encountered an error communicating with an Internet Gateway Device.  Please check that the device is functioning properly. BITS will not attempt to use this device until the next system reboot. Error code: 0x80040500.
```

I tried setting UPNP_VERSION_MAJOR back to "1", created and started a test BITS task, and so far it seems to have resolved the issue.

Test procedure used:

http://blogs.technet.com/b/networking/archive/2008/08/14/how-to-test-the-bits-client-on-your-machine.aspx

Test machine is a Windows 10 desktop.  miniupnpd host is a router running Asuswrt-Merlin.

RMerl:
Additional information: this is the generated Event:

https://technet.microsoft.com/en-ca/library/cc734731(v=ws.10).aspx

miniupnp:
so with
```
#define UPNP_VERSION_MAJOR      2
#define UPNP_VERSION_MINOR      0
```
the BITS client doesn't work, and with 
```
#define UPNP_VERSION_MAJOR      1
#define UPNP_VERSION_MINOR      1
```
it works ? (no other difference)
Looks like a bug in Windows 10... UDA (UPnP Device Architecture) v2.0 is from September 2014, and Microsoft is a contributor !

Or maybe the UPnP version should not be put in XML descriptions 
```
<root xmlns="urn:schemas-upnp-org:device-1-0">
 <specVersion>
  <major>2</major>
  <minor>0</minor>
 </specVersion>
[...]
```

miniupnp:
UPnP-arch-DeviceArchitecture-v2.0.pdf - 2.3 Device description -  p50 : 
_\<specVersion\>
Required. In device templates, defines the lowest version of the architecture on which the device can be implemented. In actual UPnP devices, defines the architecture on which the device is implemented. Contains the following sub elements:
\<major\>
Required. Major version of the UPnP Device Architecture. Shall be 2 for devices implemented on a UPnP 2.0 architecture.
\<minor\>
Required. Minor version of the UPnP Device Architecture. Shall be 0 for devices implemented on a UPnP 2.0 architecture. Shall accurately reflect the version number of the UPnP Device Architecture supported by the device._

[...]

PR #168, opened 2015-12-11
report UPnP version 1.0 in XML descriptions to be consistent with IGD…

razzfazz:
…v2 specs

In the XML examples in all [IGDv2-related spec documents](http://upnp.org/specs/gw), the specVersion fields consistently use major=1, minor=0. Moreover, the [IGDv2 spec itself](http://upnp.org/specs/gw/UPnP-gw-InternetGatewayDevice-v2-Device.pdf) specifically says "For UPnP Versions 1.0 and 1.1". This should fix issue #167.


miniupnp:
The problem is that miniupnpd DOES support UPnP Device Architecture 2.0...
IGD v2 spec is from December 2010, so before UDA 2.0

I'm thinking about a compile time option to set PDA version

miniupnp:
See https://github.com/miniupnp/miniupnp/commit/9e65fbbded191aa86248a638ad33b5bf59cfc779

PR #169, opened 2015-12-11
fix Layer3Forwarding serviceId to be consistent with IGDv2 spec

razzfazz:
Per the [IGDv2 spec](http://upnp.org/specs/gw/UPnP-gw-InternetGatewayDevice-v2-Device.pdf) (see Table 1 in Section 2.2 or the XML in Section 3), the serviceId for the Layer3Forwarding service should be "L3Forwarding1".


miniupnp:
thanks !
Thats the same in IGD v1 spec

PR #172, opened 2015-12-11
<?xml> tags are only legal at the beginning of an XML document, not i…

razzfazz:
…nside other tags

Saw this error when inspecting miniupnpd with GUPnP-tools.


razzfazz:
Ah, looking at the spec, this method is supposed to return an XML document. So the proper fix is not to remove the offending tag, but to properly XML-escape the entire contents of SupportedProtocols (see section 2.4.3.1 of the DeviceProtection spec).

PR #178, opened 2015-12-12
escape XML in SupportedProtocols variable

razzfazz:
As per section 2.4.3.1 of the [DeviceProtection spec](http://upnp.org/specs/gw/UPnP-gw-DeviceProtection-v1-Service.pdf), the contents of the SupportedProtocols variable (which is itself an XML document) needs to be escaped when returned inside other XML.

### DIFF
--- a/miniupnpd/upnpsoap.c
+++ b/miniupnpd/upnpsoap.c
@@ -2031,14 +2031,14 @@ GetSupportedProtocols(struct upnphttp * h, const char * action, const char * ns)
 	char body[1024];
 	int bodylen;
 	const char * ProtocolList =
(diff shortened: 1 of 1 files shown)


miniupnp:
is it better to use XML entities or <![CDATA[   ]]>  ?
I guess CDATA is better for such a long XML document

razzfazz:
Didn't think of that; that does seem more readable.

miniupnp:
Don't hesitate to put quotation of the specification into comments in the code, It's easier to recall why things are done.
miniupnpd has nearly 10years now, there is a lot of code I don't remember why it is there :)

miniupnp:
see f076f368bdb868353cbdaa60b54fbbcef3e5c942

PR #180, opened 2015-12-12
fix action arguments for DeviceProtection service

razzfazz:
The definitions for the three actions for the DeviceProtection service are missing the parameters defined in [the spec](http://upnp.org/specs/gw/UPnP-gw-DeviceProtection-v1-Service.pdf). Also, the tags in the responses are incorrect (bogus "New" prefix).


razzfazz:
Verified with "Intel Device Spy for UPnP Technologies" that GetAssignedRoles and GetSupportedProtocols now work as expected with this change.

PR #184, opened 2015-12-21
Avoiding conditional directives that split up parts of statements.

RomeroMalaquias:


RomeroMalaquias:
Hello?

miniupnp:
sorry.
I don't think your commit adds much clarity.
maybe something like
```
	if(
#ifdef ENABLE_NATPMP
        !GETFLAG(ENABLENATPMPMASK) && !GETFLAG(ENABLEUPNPMASK)
#else
        !GETFLAG(ENABLEUPNPMASK)
#endif
       ) {
```
is more understandable.

PR #201, opened 2016-07-23
A few fixes for miniupnpd on DragonFly BSD

y0n3t4n1:
- fix a few build issues
- when built with ENABLE_IPV6, miniupnpd didn't respond to requests made to its IPv4 addresses.

### DIFF
--- a/miniupnpd/genconfig.sh
+++ b/miniupnpd/genconfig.sh
@@ -234,6 +234,9 @@ case $OS_NAME in
 		fi
 		echo "#define USE_IFACEWATCHER 1" >> ${CONFIGFILE}
 		echo "#define PFRULE_INOUT_COUNTS" >> ${CONFIGFILE}
+		# net.inet6.ip6.v6only has been on by default for many years
+		# and this sysctl node has been removed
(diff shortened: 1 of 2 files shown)


miniupnpd/genconfig.sh
```diff
@@ -233,6 +233,7 @@ case $OS_NAME in
 			FW=pf
 		fi
 		echo "#define USE_IFACEWATCHER 1" >> ${CONFIGFILE}
+		echo "#define PFRULE_INOUT_COUNTS" >> ${CONFIGFILE}
```
miniupnp: shouldn't the DragonFly version be tested ?
was done for OpenBSD : 
https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/genconfig.sh#L144

miniupnp:
your PR is interesting, unfortunately I'm away, I cannot merge as fast as I wish

miniupnpd/genconfig.sh
y0n3t4n1: I could, but DragonFly BSD 2.8 was released almost 6 years ago, and I can't really check if miniupnpd builds or run properly on such an old system.

PR #202, opened 2016-07-23
Fix for a number of issues found during static code analysis:

irwir:
1) mismatching types;
2) unused header files;
3) unused variables;
4) unused assigned values;
5) free() needs no check for null pointer

This relates to the **miniupnpc** library only.


miniupnpc/miniupnpc.c
```diff
@@ -432,7 +432,7 @@ build_absolute_url(const char * baseurl, const char * descURL,
 	base = (baseurl[0] == '\0') ? descURL : baseurl;
 	n = strlen(base);
 	if(n > 7) {
-		p = strchr(base + 7, '/');
+		p = (char *)strchr(base + 7, '/');
```
y0n3t4n1: `strchr()` and `strstr()` return `char *`, so you shouldn't need the cast above.

miniupnpc/upnpcommands.c
```diff
@@ -31,11 +30,11 @@ UPNP_GetTotalBytesSent(const char * controlURL,
 	char * p;
 	if(!(buffer = simpleUPnPcommand(-1, controlURL, servicetype,
 	                                "GetTotalBytesSent", 0, &bufsize))) {
-		return UPNPCOMMAND_HTTP_ERROR;
+		return (UNSIGNED_INTEGER)UPNPCOMMAND_HTTP_ERROR;
```
y0n3t4n1: you can move the cast to the macro definition so as you don't need to sprinkle everywhere it's used.

irwir:
Thanks, **y0n3t4n1**
Obviously, there is no need to cast to character pointer.
That is odd, I vaguely remember seeing something about pointers to void. Possibly I copy-pasted that from one of my previous versions without proper examination.

But idea to move type cast to macro definition cannot be used, because there are functions returning the same error code as signed value.

miniupnpc/miniupnpc.c
irwir: Now I recalled that this cast came from VS2012, where `strchr()` was returning `const char *`.

miniupnpc/miniupnpc.c
```diff
@@ -706,13 +706,13 @@ UPNP_GetIGDFromUrl(const char * rootdescurl,
 	int descXMLsize = 0;
 
 	descXML = miniwget_getaddr(rootdescurl, &descXMLsize,
-	                           lanaddr, lanaddrlen, 0, NULL);
+	   	                       lanaddr, lanaddrlen, 0, NULL);
```
iphydf: This violates the coding standard dictated by surrounding code. It is also not a semantic change.

iphydf:
@irwir which tool did you use for static analysis?

PR #205, opened 2016-07-29
Set _GNU_SOURCE in asyncsendto.c

ffontaine:
Add #define on _GNU_SOURCE at the very beginning of asyncsendto.c file
as this file uses in6_pktinfo which is protected by __USE_GNU in
libc/inet/netinet/in.h. Currently, this flag is set by the Makefile in
CFLAGS however CFLAGS could be overwritten by the build system.

Signed-off-by: Fabrice Fontaine <fabrice.fontaine@orange.com>

### DIFF
--- a/minissdpd/asyncsendto.c
+++ b/minissdpd/asyncsendto.c
@@ -5,6 +5,9 @@
  * This software is subject to the conditions detailed
  * in the LICENCE file provided within the distribution */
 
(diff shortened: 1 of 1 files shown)


miniupnp:
_GNU_SOURCE is set in Makefile (both minissdpd/Makefile and miniupnpd/Makefile.linux)

PR #208, opened 2016-08-29
Update testasync.c

ztecxb:
test can write or not


miniupnp:
what are you trying to update ?

PR #211, opened 2016-09-22
miniupnpc : exit status according to upnp status

yanngarras:
Aims to fix issue 210.

The goal is that miniupnp exit status gives information on error or success during operation.
Now it will 0 if everything went fine, and other int if error occurred.


miniupnp:
oops I changed the code before seeing this PR

yanngarras:
no problem. Could you return a code in RemoveRedirect
and RemoveRedirectRange too please ?

Thanks

2016-09-22 22:03 GMT+02:00 Thomas BERNARD <notifications@github.com>:

PR #243, opened 2017-06-11
Mainly adding fd_set related header missing

devnexen:


miniupnp:
on which building setup is that needed ?

devnexen:
Tried to compile on openbsd

PR #258, opened 2017-09-16
Fixed incompatible types (from 'int *' to 'const char *') warning

raspopov:
VC++ 2015: miniupnpc\minissdpc.c(619): warning C4133: 'function': incompatible types - from 'int *' to 'const char *'


miniupnp:
Isn't it a Windows only patch ?
```
 int
     setsockopt(int s, int level, int optname, const void *optval,
         socklen_t optlen);
```

raspopov:
Oh, you are right! Need to add "#ifdef _WIN32" around just like at next line below.

Issue #277, opened 2018-01-19
allow (mis?)reporting IGDv1 to fool broken igd clients


yangfl:
Forwarded from https://bugs.debian.org/885960

Many IGD clients (most notably MS BITS) look for IGD v1 and do not recognize v2 as a valid igd service

Since v2 is backwards compatible with v1, perhaps there should be a way to force miniupnpd to report v1

miniupnp:
this is a know issue. see #167 
Default has been reverted to UDA 1.1 for 2 years now : 58b130116ba7f1a3669cd822811822b4967c5ac7

Please use latest code.

nyetwurk:
I am still seeing urn:schemas-upnp-org:device:InternetGatewayDevice:2 on 2.0

miniupnp:
@nyetwurk  what is in your config.h file ?

nyetwurk:
```
/* UPnP version reported in XML descriptions
 * 1.0 / 1.1 / 2.0 depending on which UDA (UPnP Device Architecture) Version */
#define UPNP_VERSION_MAJOR      1
#define UPNP_VERSION_MINOR      1
#define UPNP_VERSION_MAJOR_STR  XSTR(UPNP_VERSION_MAJOR)
#define UPNP_VERSION_MINOR_STR  XSTR(UPNP_VERSION_MINOR)
```

nyetwurk:
```
/* Enable the support of IGD v2 specification.
 * This is not fully tested yet and can cause incompatibilities with some
 * control points, so enable with care. */
#define IGD_V2

#ifdef IGD_V2
/* Enable DeviceProtection service (IGDv2) */
#define ENABLE_DP_SERVICE
/*#define ENABLE_HTTPS*/
/*#define HTTPS_CERTFILE "/path/to/certificate.pem"*/
/*#define HTTPS_KEYFILE "/path/to/private.key"*/

/* Enable WANIPv6FirewallControl service (IGDv2). needs IPv6 */
#ifdef ENABLE_IPV6
#define ENABLE_6FC_SERVICE
#endif /* ENABLE_IPV6 */
#endif /* IGD_V2 */
```

miniupnp:
So you have configured UPNP version 1.1 (UDA 1.1)
but enabled IGD 2 (if it is disabled it will be IGD 1)

miniupnp:
[...]

PR #282, opened 2018-02-20
Allow runtime override of igd to v1 for people running binaries with v2 enabled

nyetwurk:
Towards miniupnp/miniupnp#277

Hopefully this allows people running distros that blindly enable v2 to override it at run time.

tested with windows 10 and windows 7 upnp clients


miniupnp:
see 80779ff4f77b5717b93d5099f421b1ea2636364d

PR #296, opened 2018-04-27
Modernize and cleanup CMakeLists.txt

btolfa:
[Effective Modern CMake](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1)

- Set the minimum required version for CMake to 3.5
- Added `UPNPC_BUILD_SAMPLE` option for building sample client application (upnpc.c)
- Added interface target library `miniupnpc-private` for common settings like compile definitions, flags and include directories. It helps to keep global scope clean.
  https://cmake.org/cmake/help/v3.11/manual/cmake-buildsystem.7.html#interface-libraries
- Replaced `-fPIC` compile option with `POSITION_INDEPENDENT_CODE` target property
  https://cmake.org/cmake/help/v3.11/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE
- Set usage requirements for `libminiupnpc-*` targets for more convenient usage through `add_subdirectory()` (as cmake subproject)
  https://cmake.org/cmake/help/v3.11/manual/cmake-buildsystem.7.html#build-specification-and-usage-requirements
- Removed old and outdated code from CMakeLists.txt

### DIFF
--- a/miniupnpc/CMakeLists.txt
+++ b/miniupnpc/CMakeLists.txt
@@ -1,73 +1,56 @@
-cmake_minimum_required (VERSION 2.6)
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 project (miniupnpc C)
+
 set (MINIUPNPC_VERSION 2.0)
 set (MINIUPNPC_API_VERSION 17)
 
-if (NOT CMAKE_BUILD_TYPE)
-  if (WIN32)
-    set (DEFAULT_BUILD_TYPE MinSizeRel)
-  else (WIN32)
-    set (DEFAULT_BUILD_TYPE RelWithDebInfo)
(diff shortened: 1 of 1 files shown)


miniupnp:
isn't it possible to support older versions of CMake without too much modifications ?

anonimal:
>isn't it possible to support older versions of CMake without too much modifications ?

Is there a target version that miniupnp is aiming to support? Most commonly used systems these days (I haven't confirmed with amiga or solaris) [use 3.5 minimum](https://build.getmonero.org/waterfall).

miniupnp:
@anonimal I have an old debian jessie with 3.0.2...
I was wondering which features you use requires 3.5

Issue #298, opened 2018-05-07
GetExternalIPAddress returns NAT-ed address from private IPv4 block


pali:
It is very often that user's router with miniupnpd is already behind NAT.

Lot of ISPs provides public IPv4 address to end user just via NAT 1:1 mechanism. Which means that DHCPv4 server assigns for user router one private IPv4 address and ISP configure his infrastructure to translate user's public IPv4 address to that private IPv4 address assigned by DHCPv4 server (it is like forwarding all TCP and UDP ports).

So in reality user has his own public IPv4 address (in some cases it is just dynamic and may change in next day or weeks), but with NAT 1:1.

And in this case miniupnpd function GetExternalIPAddress() returns that private IPv4 address instead of correct public. Software which uses UPnP protocol for getting public IPv4 address does not work.

I would suggest to extend miniupnpd functionality to check if ip address on external interface is really public IPv4 (not in 10.0.0.0/8, 192.168.0.0/24, ... blocks) and configuration option to learn correct public IPv4 address via e.g. STUN protocol from STUN server. Via STUN protocol it is also possible to detect if user has above NAT 1:1 or symmetric NAT.

Maybe in case when symmetric NAT is detected miniupnpd should reject any request for ip address forwarding and return error (as it would not work).

pali:
@miniupnp ISP forwards all tcp and udp ports from public ip address to private address assigned by dhcp. So there is no NAT traversal, it is NAT 1:1.

This setup is very common and the only thing which is needed is to learn correct public ip address.

miniupnp:
what's the point in using NAT 1:1 for an ISP ?
usually ISP which have not enough public IP address available use RFC1918 address for their customers and use NAT to map (for example) 256 public IP addresses on 256*16=4096 customers or whatever.

In the case of NAT 1:1, why don't you use the `ext_ip=` option in miniupnpd.conf?

pali:
> what's the point in using NAT 1:1 for an ISP ?

I'm not ISP, so I do not know exactly. I think that it simplify configuration for ISP's DHCP server which assigns IP addresses when ISP has customers with and also without public IP address. For customer which ask for public IP address, ISP just add a new record to their database and do not need to change routing nor how DHCP server assign IP addresses.

[...]

PR #307, opened 2018-05-19
miniupnpd: Learn public IP address and detect NAT via STUN protocol

pali:
Normally port forwarding works only in case miniupnpd have public IP address and it is not running on network hidden behind NAT.

There is one exception: when miniupnpd host has private IP address with unrestricted NAT 1:1 -- i.e. it private IP address has 1:1 mapping with some public IP address and all incoming packets to public IP address are NAT-ed to private IP address without any filtering or dropping. In some countries this is common behavior of ISPs how to provide public external IP address.

This pull request implements STUN client protocol for learning external IP address and NAT type via external STUN server. This feature is disabled by default and user can enable it via config options `ext_perform_stun`, `ext_stun_host` and `ext_stun_port`.

When sending STUN binding requests we need to enable incoming UDP traffic to selected ports as responses to sent packets comes from different IP addresses/ports as to which were sent. This is how STUN protocol is working and tests if NAT allows and passes inbound connections.

When miniupnpd detects that IP address on external interface was changed (either by watching for notifications or when user instruct upnpnd via SIGUSR1 signal), then miniupnpd perform STUN requests again to learn new external IP address and NAT type. All this happens only in case config option `ext_perform_stun` is enabled (disabled by default).

In case miniupnpd detects that is behind restrictive NAT (all types expects above NAT 1:1) or upstream router does some filtering of incoming packets which leads to dropping packets on forwarded ports, then miniupnpd rejects all port forwarding requests from UPnP IGD/PCP clients (as port forwarding in this case does not work).

Also when clients ask for public IP address then miniupnpd checks that it really knows external public IP address. And if not, then it correctly send error message instead of sending incorrect internal / private IP address.

See also discussion: https://github.com/miniupnp/miniupnp/issues/298

### DIFF
--- a/miniupnpd/upnpstun.c
+++ b/miniupnpd/upnpstun.c
@@ -1,0 +1,464 @@
+/* MiniUPnP project
+ * http://miniupnp.free.fr/ or http://miniupnp.tuxfamily.org/
+ * (c) 2018 Pali Rohár
+ * This software is subject to the conditions detailed
+ * in the LICENCE file provided within the distribution */
+
+#include <sys/select.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <unistd.h>
+#include <netdb.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#ifndef TEST_LINUX_DEBUG_APP
+#include "config.h"
+#endif
+
+#include "upnpstun.h"
+
+#if defined(USE_NETFILTER)
+#include "netfilter/iptcrdr.h"
+#endif
+#if defined(USE_PF)
+#include "pf/obsdrdr.h"
+#endif
+#if defined(USE_IPF)
+#include "ipf/ipfrdr.h"
+#endif
+#if defined(USE_IPFW)
+#include "ipfw/ipfwrdr.h"
+#endif
+
+#ifdef TEST_LINUX_DEBUG_APP
+static int add_filter_rule2(const char *ifname, const char *rhost, const char *iaddr, unsigned short eport, unsigned short iport, int proto, const char *desc);
+static int delete_filter_rule(const char * ifname, unsigned short port, int proto);
+#endif
+
+/* Generate random STUN Transaction Id */
+static void generate_transaction_id(unsigned char transaction_id[12])
+{
+	int i;
+
+	for (i = 0; i < 12; ++i)
+		transaction_id[i] = random()%255;
+}
+
+/* Create and fill STUN Binding Request */
+static void fill_request(unsigned char buffer[28], int change_ip, int change_port)
+{
+	/* Type: Binding Request */
+	buffer[0] = 0x00;
+	buffer[1] = 0x01;
+
+	/* Length: One 8-byte attribute */
+	buffer[2] = 0x00;
+	buffer[3] = 0x08;
+
+	/* Magic Cookie: 0x2120A442 */
+	buffer[4] = 0x21;
+	buffer[5] = 0x12;
+	buffer[6] = 0xA4;
+	buffer[7] = 0x42;
+
+	/* Transaction Id */
+	generate_transaction_id(buffer+8);
+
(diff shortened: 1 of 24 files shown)


pali:
Thanks for review! I created a new pull request https://github.com/miniupnp/miniupnp/pull/314 for your -o option change https://github.com/miniupnp/miniupnp/commit/8bc6d6f5564cf64e99f1c2cf46022f9221af123f

miniupnp:
@pali: can you answer the question about` random()%255;`

miniupnpd/upnpstun.c
```diff
@@ -0,0 +1,464 @@
...
+
+	if (argc != 3 && argc != 2) {
+		printf("Usage: %s stun_host [stun_port]\n", argv[0]);
+		return 1;
+	}
+
+	if (argc == 2)
+		argv[2] = "0";
+
+	srandom(time(NULL) * getpid());
```
miniupnp: srandom() is called in miniupnpd.c somewhere.

pali:
Which question? (I do not see any there)

`random()%255` is used only in `generate_transaction_id` function. And it is because https://tools.ietf.org/html/rfc5389#section-3 describe it as: `Both types of transactions include a transaction ID, which is a randomly selected 96-bit number.`

Hm.. probably it should be `random()%256` to generate number `0..0xFF`.

miniupnpd/upnpstun.c
miniupnp: oops sorry it is in the main() there :)

miniupnp:
sorry it looks like I needed to "submit" the review comments..

miniupnpd/upnpstun.c
```diff
@@ -0,0 +1,464 @@
...
+static int delete_filter_rule(const char * ifname, unsigned short port, int proto);
+#endif
+
+/* Generate random STUN Transaction Id */
+static void generate_transaction_id(unsigned char transaction_id[12])
+{
+	int i;
+
+	for (i = 0; i < 12; ++i)
+		transaction_id[i] = random()%255;
```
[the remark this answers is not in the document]
pali: Yes, %256 or &255 is correct.

miniupnpd/upnpstun.c
pali: It is undef TEST_LINUX_DEBUG_APP which produce testing stun client application. Useful for development and debugging.

alexus-m:
@pali:
> When sending STUN binding requests we need to enable incoming UDP traffic to selected ports as responses to sent packets comes from different IP addresses/ports as to which were sent. This is how STUN protocol is working and tests if NAT allows and passes inbound connections.

What is a list of IP-addresses and UDP-ports expected for Binding Responses?
(Sorry I'm not able to extract it from code)

Let me explain an outcome of my test scenario:
I'm OpenWRT user with installed miniupnpd v2.1.20180706-1.
I've configured stun.onesuite.com:3478 as a STUN server.
In order to get a positive answer
"daemon.info miniupnpd[26312]: STUN: ext interface pppoe-wan has now public IP address 178.45.123.456: Port forwarding is now enabled"
I need to configure firewall with
config rule
	option target 'ACCEPT'
	option src 'wan'
	option proto 'udp'
	option name 'STUN response'
	option src_ip '64.77.226.37 64.77.226.17'
	option src_port '3478 3479'
Otherwise I get:
daemon.warn miniupnpd[26182]: STUN: ext interface pppoe-wan with IP address 178.45.123.456 is now behind restrictive NAT with public IP address 178.45.123.456: Port forwarding is now impossible

pali:
> What is a list of IP-addresses and UDP-ports expected for Binding Responses?

Target address and port of Binding response is same as source address/port of Binding request.

Source IP address of binding request is IP address of local interface based on routing tables. Source port is random. Destination address and destination port is configured by ext_stun_host and ext_stun_port.

Upnpd automatically allow communitation to the source port of binding request (and therefore target port of binding response). After stun processing is done, ports are removed from allowed list. Look for add_filter_rule2 and delete_filter_rule function calls in miniupnpd/upnpstun.c file for details.

PR #311, opened 2018-06-19
A bunch of small changes

irwir:
That is what was left after merging changes for 64-bit Windows (you did a good job then).
This PR is supposed to decrease the number of warnings and slightly improve code style.

General considerations.

- Assignments of `NULL` to local variable after `free()` were disliked by static analysers.
The reason was: assigned value was never used.
Also compilers might optimize it away.
Therefore these assignment add very little to code safety, and could be dropped.
If you still would like to keep it, a simple macro might be used to eliminate these assignments in Release configurations.
- Use of the same parameter names in function declaration and definition should improve coding style.
- Enforced type consistency eliminates warnings (and prevents errors).

Additional comments.

`connecthostport.c`
Missed replacement: `-1` should have been `INVALID_SOCKET`.
`ai_addrlen` field is declared as `size_t` in MS VC, but parameter is `int` (truncation warning fix).

`minisoap.c`
Declaration and initialization in one line.

`minissdpc.c`
Prefix `dw` stands for `DWORD`, which is defined as `unsigned long`; and the correct format should be `%lu`, not `%ld`

`minissdpc.h`
Parameter names changed to match function's definitions.

`miniupnpc.c`
`strlen()` returns value of `size_t` type in MS VC. Hence the change from `int l, n;` to `size_t`
In `UPNP_GetValidIGD` checks were removed because some conditions should be always true.
Quick explanation:
`devlist` is not empty (line 581), therefore `ndev` is at least 1 in line 591
`desc` cannot be `NULL` after lines 594-595

`miniwget.c`
The first `i<n` condition was already verified in the loop header.
Change `int l;` to `size_t ;l` because of pointer arithmetics and sizeof.

`minixml.h`
Parentheses around macros and its parameters as a standard safety measure (static analysis).

upnpcommands.c
The constant `UPNPCOMMAND_HTTP_ERROR` is defined as signed, but it had to be used both as signed and unsigned (type mismatch warning fix).

### DIFF
--- a/miniupnpc/connecthostport.c
+++ b/miniupnpc/connecthostport.c
@@ -183,7 +183,7 @@ SOCKET connecthostport(const char * host, unsigned short port,
 #endif
 		return INVALID_SOCKET;
 	}
-	s = -1;
+	s = INVALID_SOCKET;
 	for(p = ai; p; p = p->ai_next)
 	{
 		s = socket(p->ai_family, p->ai_socktype, p->ai_protocol);
@@ -208,7 +208,7 @@ SOCKET connecthostport(const char * host, unsigned short port,
 			PRINT_SOCKET_ERROR("setsockopt");
 		}
 #endif /* #ifdef MINIUPNPC_SET_SOCKET_TIMEOUT */
-		n = connect(s, p->ai_addr, p->ai_addrlen);
+		n = connect(s, p->ai_addr, (int)p->ai_addrlen);
 #ifdef MINIUPNPC_IGNORE_EINTR
 		/* EINTR The system call was interrupted by a signal that was caught
 		 * EINPROGRESS The socket is nonblocking and the connection cannot
@@ -261,4 +261,3 @@ SOCKET connecthostport(const char * host, unsigned short port,
 #endif /* #ifdef USE_GETHOSTBYNAME */
 	return s;
 }
-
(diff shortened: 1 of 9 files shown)


miniupnpc/connecthostport.c
```diff
@@ -208,7 +208,7 @@ SOCKET connecthostport(const char * host, unsigned short port,
 			PRINT_SOCKET_ERROR("setsockopt");
 		}
 #endif /* #ifdef MINIUPNPC_SET_SOCKET_TIMEOUT */
-		n = connect(s, p->ai_addr, p->ai_addrlen);
+		n = connect(s, p->ai_addr, (int)p->ai_addrlen);
```
miniupnp:
```
     int
     connect(int s, const struct sockaddr *name, socklen_t namelen);
```
So I'm not sure casting to int is a good idea

irwir: That kind of issues is unfortunately common in socket code: [connect](https://msdn.microsoft.com/en-us/library/windows/desktop/ms737625(v=vs.85).aspx)  function in Microsoft library uses `int`.

A workaround could be taken [here](https://github.com/ARMmbed/mbedtls/blob/8266acacc8d6e1c65fba9a048f56339d0827b2fe/library/net_sockets.c#L92).
If you need, an additional commit could be made.

PR #312, opened 2018-07-01
miniupnpd: support newer libnftnl versions

Lochnair:
Trying my hand at getting nftables support up to speed.

In libnftnl versions >=1.0.7 a lot of [aliases and compat were removed](https://git.netfilter.org/libnftnl/commit/?id=907a9f8e5a93f5bcd449643eb3916a656d634758) breaking miniupnpd's nftables support completely. I've fixed all the errors caused by this I'm aware of. Not sure if there's some warnings too that should be fixed.

What's currently missing is writing the update_portmapping and update_portmapping_desc_timestamp functions for the nftables backend as done for others in bdac007. Though I doubt I've got the code-fu to do that.

### DIFF
--- a/miniupnpd/Makefile.linux_nft
+++ b/miniupnpd/Makefile.linux_nft
@@ -74,6 +74,13 @@ endif # ($(TEST),1)
 
 LDLIBS += $(shell $(PKG_CONFIG) --static --libs-only-l libssl)
 
+TEST := $(shell $(PKG_CONFIG) --exists uuid && echo 1)
+ifeq ($(TEST),1)
+LDLIBS += $(shell $(PKG_CONFIG) --static --libs-only-l uuid)
+else
+$(info please install uuid-dev package / libuuid)
+endif # ($(TEST),1)
+
 TESTUPNPDESCGENOBJS = testupnpdescgen.o upnpdescgen.o
 
 EXECUTABLES = miniupnpd testupnpdescgen testgetifstats \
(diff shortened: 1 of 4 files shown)


miniupnp:
I need to add linux nft build to travis-ci to be able to check that everything is ok

miniupnp:
@Lochnair does it still build with older libnftnl ?

Lochnair:
@miniupnp Just tried compiling against libnftnl 1.0.5 on Alpine 3.3, which gets me as far as on newer versions. The new functions were introduced in 1.0.4, so anyone using a >= 1.0.4 should be fine, which includes every distro I've checked so far.

Lochnair:
I've tried to write the update_portmapping and update_portmapping_desc_timestamp functions for nft in second commit.  It compiles in both my Arch environment and on the Alpine 3.3 container, but I haven't had the chance to actually test if it runs properly yet.

miniupnp:
travis-ci is not launching builds on the project...
https://travis-ci.org/miniupnp/miniupnp

miniupnp:
And I have merged but it doesn't appears here...

miniupnp:
I have set up a clone of this github project on framagit.org
https://framagit.org/miniupnp/miniupnp/-/jobs/183109

```
upnpstun.o: In function `perform_stun':
/builds/miniupnp/miniupnp/miniupnpd/upnpstun.c:363: undefined reference to `delete_filter_rule'
collect2: error: ld returned 1 exit status
```

PR #348, opened 2019-02-06
minissdpc: add a libuv client

alkino:
Here is a quick&dirty but working implementation of the minissdp client with libuv to have an asynchronous client.

What do you think about it? Mergeable? Useful?

### DIFF
--- a/miniupnpc-libuv/README
+++ b/miniupnpc-libuv/README
@@ -1,0 +1,11 @@
+Here is a basic libuv implementation of minissdp client.
+
+You can find an example find attached to test it.
+
+BUILD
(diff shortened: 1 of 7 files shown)


miniupnp:
please add a README file in the directory to explain

alkino:
Do you mean I should put it in a single directory near libevent?

miniupnp:
> Do you mean I should put it in a single directory near libevent?
>
Well you just add two files without explaining anything.
How to use them ? why ?

alkino:
Here is an update in a separated dir with a README and an example usage file.

PR #362, opened 2019-04-09
Fix issues from static analysis and simplify.

irwir:
**minissdpc.c**
- warnings for redeclared variable `p` in lines 524 and 530 (hides the previous declaration)
- there is no reason to use platform-specific heap; standard C library functions `malloc`/`free` are fine
- part of condition always true in line 572 - `(dwRetVal == ERROR_BUFFER_OVERFLOW)` 

**miniupnp.c**
- condition always true in line 348 for `WIN32` - `(devlist == NULL)`
- `memcmp` might access memory beyond the end of string in lines 429 and 462

**upnpcommands.c**
- restrict scope of variable `p`


miniupnp:
thank you for the report. I don't think I will merge as-is

irwir:
Changed code according to the latest commit.

PR #366, opened 2019-04-21
CMake: Add listdevices

neheb:
Matches the Makefile


miniupnp:
merged.
82b117ee7f7a471f1f833ffb5b5b980cd44a97f0

prissi:
But now Cmake adds listdevices to the lib, where it overlays the main program's "_main" . It certainly does not belong into the general list of sources, it is just needed for building listdevices. So this patch breaks linking the resulting a static library libminiupnpc.a for any program.

PR #367, opened 2019-04-25
Fixing the Upnpc delivery process for windows targets

NathanRichard:
# Changes
I reworked the Appveyor build process and introduced the following changes :
- The python module is built for Python 2.7 (as before), Python 3.5 and Python 3.7. Adding support for new python versions should be relatively easy.
- The python module is distributed both as a raw egg file (which existed previously but was buried in a zip file with other non-required binary files), and a wheel file (which seems to be the recommended distribution method for python modules now).

# Remarks
- I only tested the .egg and .whl files for Python 3.7 under Windows 7 32 bits, which is the environment I was interested in. The other builds completed without errors but may require further validation.
- One of the main fixes that allowed the Python 3 versions to build was to remove the --compiler directive given to setuptool : it then naturally used the MSVC compiler to build the module as it is the default compiler for modern python versions and it seems mingw32 does not play well with these versions. However I think the C part of the project was not affected and is still built using mingw32. So this means the MVSC-built python module is linking against the mingw-built upnpc libraries. I am actually surprised its works so well in practice (we may have to thank the strong C-ABI for that), but I am not whether it is by chance or it is a guaranty of the standard. Comments and insights are very welcome on that aspect.
- References to mingw are all over the project, in comments, config files etc. I did not touch those, but it would probably be worth checking if all these references are still valid since we now also use MSVC.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,8 @@
-version: '2.1.{build}'
+environment:
+  APP_VERSION: '2.1'
+
+
+version: '$(APP_VERSION).{build}'
 
 image: Visual Studio 2017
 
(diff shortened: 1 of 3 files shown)


miniupnp:
using mingw or MSVC should produce compatible code.
The only difference would be in library dependencies

We could add a nmake makefile for building with "cl"
or use the MSVC project files : https://github.com/miniupnp/miniupnp/tree/master/miniupnpc/msvc

miniupnp:
we could also modify setupmingw32.py (or create a new setupwin32.py) to drop dependency on libminiupnpc.a and add all *.c

Issue #380, opened 2019-06-20
Nftables in miniupnpd (latest revision from git)


rtorrentuser:
When I start uTorrent (which uses UPNP) rules are added to nftables as follows:

table ip filter {
	chain MINIUPNPD {
		iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol udp accept
		iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol tcp accept
	}
}
table ip nat {
	chain MINIUPNPD {
		iif "br1" udp dport 16422 dnat to 192.168.0.4:16422
		iif "br1" tcp dport 16422 dnat to 192.168.0.4:16422
	}
}

Upon program exit only rules in 'table ip nat' are deleted and those in 'table ip filter' remain "forever":

table ip filter {
	chain MINIUPNPD {
		iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol udp accept
		iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol tcp accept
	}
}
table ip nat {
	chain MINIUPNPD {
	}
}

I think that this is a wrong behavior...

miniupnp:
@svenauhagen any idea ?

svenauhagen:
@miniupnp general question, does miniupnpd flush the chains on exit or do I have to use the scripts?

miniupnp:
miniupnpd does not flush anything on exit !

the problem here is that the nat rules have been removed, but not the filter one.
I think it is utorrent who asked to remove the port mapping when exiting

rtorrentuser:
Sorry, my description of  the problem was incomplete...  The main point is not that the filter rules are not removed on program exit. The problem is that absolutely the same filter rules are added on each program start to already existing in the table. So eventually we see numerous identical records:

table ip filter {
chain MINIUPNPD {
iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol udp accept
iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol tcp accept
iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol udp accept
iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol tcp accept
iif "br1" @th,16,16 16422 ip daddr 192.168.0.4 ip protocol udp accept
[...]

PR #381, opened 2019-06-25
NFTables fixes and scripts

svenauhagen:
This commit fixes the list detection and uses the inet chain for ipv4.
The scripts got reworked as well and a display script was added.


miniupnp:
fixes #380

PR #390, opened 2019-07-29
Fix error code for 708

psibi:
According to the specification
http://upnp.org/specs/gw/UPnP-gw-WANIPConnection-v2-Service.pdf, the
708 error code corresponds to InvalidLayer2Address. Also
WildcardNotPermittedInSrcIP is for 715 which is already properly mapped.


miniupnp:
thank you for spotting this !

PR #395, opened 2019-09-01
Rework nft_removeall.sh to preserve other nftables structures

paul-chambers:
Rework nft_removeall.sh to preserve nftables structures that miniupnpd didn't add. Important for firewalld and sshguard co-existance.


miniupnp:
Thanks.
Has it been tested ? (I'm not able to test myself)

paul-chambers:
Yes, on my firewall machine (running Debian Buster). I'm using firewalld with the nftable backend, and was running into problems when I started and stopped miniupnpd. With these changes, firewalld is surviving intact across miniupnpd restarts.

PR #403, opened 2019-09-28
Update case statements to fix compiler 'fall through' warnings

paul-chambers:
A couple seem intentional (linux/ifacewatcher.c). Most look like they were unintentionally omitted (netfilter_nft/nftnlrdr_misc.c).

Resubmitted without unintended indentation changes included in previous pull request.


paul-chambers:
--firewall= and Makefile tweaks are to make it easier to compile the iptables version after making changes that should only affect nftables. I can't easily test the iptables version, but at least I can check it compiles without warnings.

miniupnp:
thanks !

Issue #397, opened 2019-09-11
Miniupnpd + nftables does not seem to work properly


4np:
Ever since I switched to `nftables`, I have been unable to get `miniupnpnd` to work. I have been running it for years before without issues using iptables / netfilter.

- Distro: [Gentoo Linux](https://gentoo.org)
- Kernel: Linux [4.19.66-gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/gentoo-sources/gentoo-sources-4.19.66.ebuild)
- miniupnpd version: [2.1.20190902](https://gentoo.zugaina.org/net-misc/miniupnpd)
- two nics:
  - enp2s0 is LAN
  - eno0 is WAN

The base ruleset I have configured is as follows:

```bash
table ip filter {
	chain input {
		type filter hook input priority 0; policy accept;
		ct state established,related accept
		ip protocol icmp counter packets 146 bytes 17104 drop
		tcp dport { ssh, http, https } ct state new counter packets 74 bytes 3944 accept
		iifname "lo" accept
		iifname "enp2s0" counter packets 3792 bytes 1012502 accept
		iifname "eno0" counter packets 996 bytes 93929 drop
		drop
	}

	chain output {
		type filter hook output priority 100; policy accept;
	}

	chain forward {
		type filter hook forward priority 0; policy drop;
		iifname "enp2s0" oifname "eno0" accept
		iifname "eno0" oifname "enp2s0" ct state established,related accept
	}
}
table ip nat {
	chain prerouting {
		type nat hook prerouting priority 0; policy accept;
	}

	chain postrouting {
		type nat hook postrouting priority 100; policy accept;
		oifname "eno0" masquerade random,persistent
	}
}
```

When I start `miniupnpd`, the ruleset becomes as follows as the `miniupnpd` tables and chains are being added:

```conf
table ip filter {
	chain input {
		type filter hook input priority 0; policy accept;
		ct state established,related accept
		ip protocol icmp counter packets 147 bytes 17188 drop
		tcp dport { ssh, http, https } ct state new counter packets 82 bytes 4364 accept
		iifname "lo" accept
		iifname "enp2s0" counter packets 3980 bytes 1058250 accept
		iifname "eno0" counter packets 1017 bytes 95803 drop
		drop
	}

	chain output {
		type filter hook output priority 100; policy accept;
	}

	chain forward {
		type filter hook forward priority 0; policy drop;
[...]

PR #404, opened 2019-10-01
Significant updates to nftables support

paul-chambers:
Branch with significant changes that updates nftables support in miniupnpd to work with current versions of libnftnl (and nftables underneath). Also includes some maintainability and readability improvements.

Also tested independently by @4np. I believe this fixes issue #397

### DIFF
--- a/miniupnpd/commonrdr.h
+++ b/miniupnpd/commonrdr.h
@@ -73,7 +73,7 @@ update_portmapping_desc_timestamp(const char * ifname,
 typedef enum {
 	RDR_TABLE_NAME,
 	RDR_NAT_PREROUTING_CHAIN_NAME,
-	RDR NAT_POSTROUTING_CHAIN_NAME,
+	RDR_NAT_POSTROUTING_CHAIN_NAME,
 	RDR_FORWARD_CHAIN_NAME,
 } rdr_name_type;
 
--- a/miniupnpd/commonrdr.h
+++ b/miniupnpd/commonrdr.h
@@ -65,16 +65,22 @@ update_portmapping_desc_timestamp(const char * ifname,
                    const char * desc, unsigned int timestamp);
 
 #ifdef USE_NFTABLES
-/* only provided by nftables implementation at the moment */
+/*
+ * only provided by nftables implementation at the moment.
+ * Should be implemented for iptables too, for consistency
+ */
 
 typedef enum {
-	TABLE_NAME,
-	TABLE4_NAME,
-	TABLE6_NAME,
-	NAT_CHAIN_NAME,
-	NAT_POSTROUTING_CHAIN_NAME,
-	FORWARD_CHAIN_NAME,
+	RDR_TABLE_NAME,
+	RDR_NAT_PREROUTING_CHAIN_NAME,
+	RDR NAT_POSTROUTING_CHAIN_NAME,
+	RDR_FORWARD_CHAIN_NAME,
 } rdr_name_type;
+
+/*
+ * used by the config file parsing in the core
+ * to set
+ */
 
 int set_rdr_name( rdr_name_type param, const char * string );
 
--- a/miniupnpd/commonrdr.h
+++ b/miniupnpd/commonrdr.h
@@ -64,4 +64,20 @@ update_portmapping_desc_timestamp(const char * ifname,
                    unsigned short eport, int proto,
                    const char * desc, unsigned int timestamp);
 
+#ifdef USE_NFTABLES
+/* only provided by nftables implementation at the moment */
+
+typedef enum {
+	TABLE_NAME,
+	TABLE4_NAME,
+	TABLE6_NAME,
+	NAT_CHAIN_NAME,
+	NAT_POSTROUTING_CHAIN_NAME,
+	FORWARD_CHAIN_NAME,
+} rdr_name_type;
+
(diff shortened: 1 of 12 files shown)


paul-chambers:
Build appears to be failing because `config.h` isn't being regenerated with `USE_NFTABLES `set. Builds without error for me.

miniupnpd/netfilter_nft/nftnlrdr.c
```diff
@@ -219,11 +325,11 @@ delete_filter_rule(const char * ifname, unsigned short port, int proto)
 	struct nftnl_rule *r;
 	UNUSED(ifname);
 
-	reflesh_nft_cache_filter();
+	refresh_nft_cache_filter();
```
miniupnp: what is changed ?

miniupnp: ah ok it was a typo ;)

miniupnpd/netfilter_nft/nftnlrdr_misc.c
```diff
@@ -32,23 +32,45 @@
...
+#include <libnftnl/chain.h>
 #include <libnftnl/rule.h>
 #include <libnftnl/expr.h>
 
+#include "../commonrdr.h"
 #include "nftnlrdr_misc.h"
 #include "../macros.h"
-#include "../upnpglobalvars.h"
+
+#define DEBUG 1
```
miniupnp: that's an issue. You should build with` CFLAGS=-DDEBUG` or `#define DEBUG 1` in config.h if you want to enable debug

miniupnpd/commonrdr.h
```diff
@@ -64,4 +64,20 @@ update_portmapping_desc_timestamp(const char * ifname,
...
+/* only provided by nftables implementation at the moment */
+
+typedef enum {
+	TABLE_NAME,
+	TABLE4_NAME,
+	TABLE6_NAME,
+	NAT_CHAIN_NAME,
+	NAT_POSTROUTING_CHAIN_NAME,
+	FORWARD_CHAIN_NAME,
+} rdr_name_type;
```
miniupnp: shouldn't you use more specific names like NF_TABLE_NAME ?

miniupnpd/miniupnpd.c
```diff
@@ -1267,6 +1267,17 @@ init(int argc, char * * argv, struct runtime_vars * v)
...
+			case UPNPFORWARDCHAIN:
+				set_rdr_name(FORWARD_CHAIN_NAME, ary_options[i].value);
+				break;
+			case UPNPNATCHAIN:
+				set_rdr_name(NAT_CHAIN_NAME, ary_options[i].value);
+				break;
+			case UPNPNATPOSTCHAIN:
+				set_rdr_name(NAT_POSTROUTING_CHAIN_NAME, ary_options[i].value);
+				break;
+#else
```
miniupnp: the same modification can apply to standard iptables, right ?
I think miniupnpd_forward_chain/miniupnpd_nat_chain/miniupnpd_nat_postrouting_chain are only used in iptcrdr.c

miniupnpd/miniupnpd.c
```diff
@@ -2901,6 +2913,8 @@ main(int argc, char * * argv)
 	freeoptions();
 #endif
 
+	shutdown_redirect();
```
miniupnp: nice one !

miniupnp: thanks for spoting that !

miniupnp:
> CI build appears to be failing because `config.h` isn't being regenerated with `USE_NFTABLES `set. Builds without error for me.

that's fixed ;)

miniupnpd/netfilter_nft/nftnlrdr_misc.c
paul-chambers: Oops... sorry, that was meant to be a temporary thing, I just forgot to remove it. I've noticed I have a couple of debug messages I also meant to remove. I'll commit a fix.

miniupnpd/miniupnpd.c
paul-chambers: Yes, that's the intent. I just haven't been able to test the IPtables version, so I limited what I changed to what I could test easily. More a matter of time than anything else.

The way nat works on nftables, you must have both the prerouting and postrouting chains defined (to establish both inbound and outbound paths) irrespective of whether both chains have rules. So even though no rules are added to 'postrouting', it must be present or the rules in 'prerouting' won't work properly.
 
(part of my nefarious plan to implement privilege separation ;)

miniupnpd/commonrdr.h
paul-chambers: Good point - enums are scoped to the class that contains them in C++, but they have file scope in C - I'd forgotten that. I'll make them more unique.

miniupnpd/miniupnpd.c
paul-chambers: Didn't really matter while shutdown_redirect() was a stub. So it was a problem of my own creation :)

paul-chambers:
Thanks for the Travis fix. I think I've resolved everything.

I caught another bug in the original code. The 'description' is stored in USERDATA, a sized blob. Thus you can't just use strdup() to copy it as if it's a null-terminated string, must use strndup().

miniupnp:
merged manually

Issue #405, opened 2019-10-03
Security: All of miniupnpd runs with elevated privileges


paul-chambers:
`miniupnpd` must have elevated privileges (or at least `CAP_NET_ADMIN` and `CAP_NET_RAW` on Linux) to manipulate network rules, whatever OS it is running under.

It is common for daemons to start execution with superuser privileges, do whatever setup needs those privileges (e.g. opening/binding to ports < 1024), and then drop the elevated privileges before handling client interaction. In miniupnpd's case, it must retain some elevated privileges to be able to manipulate the system's networking rules dynamically.

From a security & robustness perspective, it is desirable to run code that handles external input (from the user or a network) with the minimum privileges possible. That way, should an attacker find an exploitable bug, they haven't gained any elevated privileges.

The common core of `miniupnpd` contains the majority of the complexity and presents the potential attack surface to the network. Thus it is highly desirable to run the core code with the minimal set of privileges required by the OS-specific network rule manipulation. Thus if the core is compromised, the attacker only gains a few elevated networking privileges, not the complete set root normally has.

miniupnp:
Please keep in mind that miniupnpd is designed to work on home routers with low memory.
So the full privilege separation with 2 process should be a build time option.

I think the first step would be to drop privileges after the init phase is done.
specifically for the nftable implementation, I don't see why the privilege are needed except during the init phase. I guess you just need to ensure that mnl_socket_open() is called during the init phase.

have a look at pf/obsdrdr.c

paul-chambers:
Point well taken. my day job is embedded Linux on platforms similar or smaller in footprint than a typical SoHo router. So I’m used to thinking ‘small’.

On Linux at least, a fork is relatively cheap. Pages are copied on write, so with a little care, the additional consumption of physical pages is minimal.

paul-chambers:
Good observation - All I need to do is move the call to `mnl_socket_open()` from doing it on-first-demand to opening it in `init_redirect`, and keep it open until `shutdown_redirect`. That means elevated privileges can be dropped after the init phase with no ill effect (since things like pid files have already been created). No need to fork at all.

Perhaps the most complicated part of such a change would be to decide what user/group to use to drop privileges :)  Ideally create a dedicated user as part of install. It's pretty common to make them parameters that can be specified in the config file, too.

I know that'll work for Linux, not so sure about other OSes, depends if it's also some form of socket interface, or privileged APIs are required.

[...]

Issue #152, opened 2015-09-15
Move to seperate repos please?


cicku:
Currently 5 projects are bundled in one repo, hard to track the changes.

Is it possible for you to create 5 repos and place each of them inside?

Thanks.

noloader:
That would be helpful.

Exaeta:
Could be helpful!

paul-chambers:
I wrote a little script to do the heavy lifting
   (https://gist.github.com/paul-chambers/cc0277b74487c41513eb9d477705f34f)
Note that this doesn't adjust any of top-level files from the original repo that are copied to all the new repos. That requires a human :)

@miniupnp does this look like the right organization? (i.e. grouping the miniupnpc projects together in one repo):

- https://github.com/paul-chambers/miniupnpc
- https://github.com/paul-chambers/miniupnpd
- https://github.com/paul-chambers/minissdpd

Neustradamus:
I confirm that to separate the code will be an excellent idea.

And after it, there will be not a problem with "Releases" and "Tags" sections like now:
- https://github.com/miniupnp/miniupnp/releases
- https://github.com/miniupnp/miniupnp/tags

PR #407, opened 2019-10-05
nft: Open a persistent connection to mnl/netlink socket at init_redirect()

paul-chambers:
Preparation for dropping elevated privileges after the initialization phase for the nftables implementation on Linux. See #405 

Move the 8k allocation on the stack for debug-printing rules in debug builds into a separate function. This keeps the stack layout consistent between debug and non-debug builds, making it easier to diagnose some classes of bugs, e.g. stack corruptions, alignment problems, etc.


miniupnpd/netfilter_nft/nftnlrdr_misc.c
```diff
@@ -82,7 +83,37 @@ static uint32_t rule_list_filter_validate = RULE_CACHE_INVALID;
...
+void
+nft_mnl_disconnect(void)
+{
+	mnl_socket_close(mnl_sock);
+	mnl_sock = NULL;
+}
+
 #ifdef DEBUG
+
+#if 0
```
miniupnp: maybe you should comment why this code is left dead ;)

miniupnp:
thanks !

miniupnp:
thank you for investing time in miniupnpd.
I'm quite busy theses days and I don't have much time to setup a router with nftables to fix and test the nftables backend

paul-chambers:
Thanks for creating miniupnp in the first place. I've been using it for years, seems only fair to contribute back when the chance presented itself :)

Perhaps once I've proven myself, I can help out more. I'd love to help separate the single repo into per-project repos for example ( #152 ). My OCD again :)

The nftables implementation could be made simpler. Not sure if the caches are really necessary, the nfnml library provides iterators that could serve the same purposes, and I'm not sure the performance gain is worth the major increase in complexity. The rule_t private structure to mirror the opaque struct nftnl_rule has to be (re)populated by multiple API calls anyhow.

miniupnpd/netfilter_nft/nftnlrdr_misc.c
paul-chambers: I just removed it, it's been dead for a while :)  I just tend to tread lightly while I'm still getting up to speed on someone else's code.

paul-chambers:
Hmm... I'm confused by the conflicts. They don't actually look like conflicts to me.

miniupnp:
I will fix the conflicts
I have the bad habit of rewriting the history ;)

miniupnp:
OK. I merged myself.
@paul-chambers I think you need to have a look at _nftpinhole.c_
please build with --igd2 and --ipv6 options

paul-chambers:
Thank you for cleaning up after me - sorry about that. The dangers of coding when tired. I'll get travis installed too.

Would it make sense to make those two options configurable by .conf, rather than at compile time?

miniupnp:
There is no reason to build theses features for them being disabled.

PR #420, opened 2020-02-17
Add Haiku Support

CodeforEvolution:
These changes shouldn't conflict with anything, and I don't expect the rename of atoui to atoui2 to affect other systems. (On Haiku, atoui was already defined in stdlib)

### DIFF
--- a/miniupnpc/CMakeLists.txt
+++ b/miniupnpc/CMakeLists.txt
@@ -96,6 +96,12 @@ if (WIN32)
 #  find_library (NSL_LIBRARY NAMES nsl)
 #  find_library (RESOLV_LIBRARY NAMES resolv)
 #  set (LDLIBS ${SOCKET_LIBRARY} ${NSL_LIBRARY} ${RESOLV_LIBRARY} ${LDLIBS})
+elseif (HAIKU)
+  target_link_libraries(miniupnpc-private INTERFACE network)
+  find_library (SOCKET_LIBRARY NAMES network)
+  find_library (NSL_LIBRARY NAMES network)
+  find_library (RESOLV_LIBRARY NAMES network)
+  set (LDLIBS ${SOCKET_LIBRARY} ${NSL_LIBRARY} ${RESOLV_LIBRARY} ${LDLIBS})  
 endif ()
 
 
--- a/miniupnpc/portlistingparse.c
+++ b/miniupnpc/portlistingparse.c
@@ -12,6 +12,10 @@
 #include "portlistingparse.h"
 #include "minixml.h"
 
+#if defined(__HAIKU__)
+#define atoui atoui2
+#endif 
+
 /* list of the elements */
 static const struct {
 	const portMappingElt code;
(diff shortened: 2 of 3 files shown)


miniupnpc/portlistingparse.c
```diff
@@ -31,7 +31,7 @@ static const struct {
 
 /* Helper function */
 static UNSIGNED_INTEGER
-atoui(const char * p, int l)
+atoui2(const char * p, int l)
```
pulkomandy: There is no atoui function in BeOS and there is no atoui function in any standard or extension I can find. So I think we should instead delete that function from Haiku?

miniupnpc/CMakeLists.txt
```diff
@@ -96,6 +96,9 @@ if (WIN32)
 #  find_library (NSL_LIBRARY NAMES nsl)
 #  find_library (RESOLV_LIBRARY NAMES resolv)
 #  set (LDLIBS ${SOCKET_LIBRARY} ${NSL_LIBRARY} ${RESOLV_LIBRARY} ${LDLIBS})
+elseif (HAIKU)
+  list(APPEND LDLIBS -lnetwork)
```
pulkomandy: It seems cleaner (I think) to do the same as the Solaris code above?

    find_library (SOCKET_LIBRARY NAMES network)
    set (LDLIBS ${SOCKET_LIBRARY})

(if hat still work, this code is commented, so maybe not)

miniupnpc/portlistingparse.c
CodeforEvolution: I think it would be best to just remove this function from Haiku. I can't find any software that uses the function directly.

miniupnpc/CMakeLists.txt
CodeforEvolution: I'll try to experiment with this.

miniupnpc/portlistingparse.c
miniupnp: even if you choose to keep atoui2() in haiku you shouldn't modify miniupnpc that way.
just add
```
#if defined(__HAIKU__)
#define atoui atoui2
#endif 
```
at the top of the file...

pulkomandy: The version in Haiku has a different prototype (unsigned atoui(const char*) so that wouldn't work

miniupnp: what do you mean ?
`#define atoui atoui2` at the top of the fie (but after the system #includes) is just like replacing every occurence of atoui by atoui2 in the remaining of the source file

CodeforEvolution: I'll give this a go when I have time.

CodeforEvolution:
Alright, those changes are all functional and the tests work too! Also, @miniupnp, do you think you may increment the version on miniupnp soon? Your last release (2.1) was in 2018, I was just curious.

CodeforEvolution:
Thank you!

PR #424, opened 2020-03-21
fix build for macos

NSBlink:
`v4` in pfvar.h is changed to `v4addr` after xnu 4570.1.46


miniupnp:
@NSBlink thank you.

I'd never thought that someone was using miniupnpd on macOS ;)

PR #425, opened 2020-03-21
fix: TypeError: a bytes-like object is required, not 'str'

iworkist:
fix: TypeError: a bytes-like object is required, not 'str'
*write function's parameter type is bytes.

### DIFF
--- a/miniupnpc/testupnpigd.py
+++ b/miniupnpc/testupnpigd.py
@@ -29,7 +29,7 @@ class handler_class(BaseHTTPRequestHandler):
 	def do_GET(self):
 		self.send_response(200)
 		self.end_headers()
-		self.wfile.write("OK MON GARS")
+		self.wfile.write(b"OK MON GARS")
 
 # create the object
 u = miniupnpc.UPnP()


miniupnp:
I'm no python expert. Is that different between python2 and python3 ?

iworkist:
> I'm no python expert. Is that different between python2 and python3 ?
yes. (I'm not python expert too. but I found the reason.)
so, this commit would work in both python 2 and 3.

please check this.
https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit
Python 3.0 uses the concepts of text and (binary) data instead of Unicode strings and 8-bit strings. All text is Unicode; however encoded Unicode is represented as binary data. The type used to hold text is str, the type used to hold data is bytes. The biggest difference with the 2.x situation is that any attempt to mix text and data in Python 3.0 raises TypeError, whereas if you were to mix Unicode and 8-bit strings in Python 2.x, it would work if the 8-bit string happened to contain only 7-bit (ASCII) bytes, but you would get UnicodeDecodeError if it contained non-ASCII values. This value-specific behavior has caused numerous sad faces over the years.

PR #438, opened 2020-04-20
miniupnpd: typo fix and update snat rules

ptpt52:
In delete_redirect_and_filter_rules() found a typo error

in update_portmapping() we have to update snat rules
and it is reasonable to ignore the failure if no snat rule found


ptpt52:
ping @miniupnp

miniupnp:
> ping @miniupnp
let me a few hours to look at it !

PR #446, opened 2020-05-08
STUN changes

pali:
This pull request fixes STUN debug/testing application and improve logging.

I tested this change with following STUN servers and it parsed all received fields:
stun.stunprotocol.org
stun.sipgate.net
stun.ekiga.net
stun.xten.com
stun.onesuite.com


pali:
http://man7.org/linux/man-pages/man3/openlog.3.html
> LOG_CONS - Write directly to the system console **if there is an error** while sending to the system logger.

So if application can connect to system logger it does **not** print message to console. And this is happening on my machine. All syslog() calls were put only into syslog and not to the stdout or console.

miniupnp:
you are right ! it is LOG_PERROR which make it print to stderr

PR #486, opened 2020-07-21
Fix Windows UWP build.

hrydgard:
Horrifying hack to get it linking on Windows UWP.

Seems the declaration of `in6addr_any` exists in Windows UWP headers but not the definition so it fails to link. So I substitute with an equivalent definition.


miniupnp (commented):
It is indeed horrifying... Are you sure the bug in UWP is not fixed yet ?

hrydgard:
Issue remains, just tested with the very latest MSVC version.

hrydgard:
Can we get this in?

miniupnp (approved):

PR #492, opened 2020-10-05
miniupnpc: By default skip invalid, disconnected and non-IGD devices

pali:
Allow to use such IGD devices by new -i option.

### DIFF
--- a/miniupnpc/upnpc.c
+++ b/miniupnpc/upnpc.c
@@ -571,6 +571,7 @@ int main(int argc, char ** argv)
 	int retcode = 0;
 	int error = 0;
 	int ipv6 = 0;
+	int ignore = 0;
 	unsigned char ttl = 2;	/* defaulting to 2 */
 	const char * description = 0;
 
@@ -624,6 +625,8 @@ int main(int argc, char ** argv)
 				description = argv[++i];
 			else if(argv[i][1] == 't')
 				ttl = (unsigned char)atoi(argv[++i]);
+			else if(argv[i][1] == 'i')
+				ignore = 1;
 			else
 			{
 				command = argv[i][1];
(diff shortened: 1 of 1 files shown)


miniupnpc/upnpc.c
```diff
@@ -705,15 +709,19 @@ int main(int argc, char ** argv)
...
 			case 3:
 				printf("UPnP device found. Is it an IGD ? : %s\n", urls.controlURL);
+				if (!ignore) goto invalid;
 				printf("Trying to continue anyway\n");
 				break;
 			default:
 				printf("Found device (igd ?) : %s\n", urls.controlURL);
+				if (!ignore) goto invalid;
 				printf("Trying to continue anyway\n");
+				break;
```
miniupnp, line 724: ?

miniupnpc/upnpc.c
```diff
@@ -705,15 +709,19 @@ int main(int argc, char ** argv)
 				break;
 			case 2:
 				printf("Found a (not connected?) IGD : %s\n", urls.controlURL);
+				if (!ignore) goto invalid;
```
miniupnp, line 712: I don't like the goto very much ;)

pali: Ok, do you have any suggestion or improvement, how to achieve this change without goto?

miniupnpc/upnpc.c
pali: Well, I added break for the last case of switch. But it is not needed.

miniupnpc/upnpc.c
miniupnp: it also introduces a memory leak

pali: Now I see it... I will think a bit more about this code.

miniupnp: see 1a1dc0f

miniupnp:
845dc469d6eaf73f3d0e497c21cebe2576d9ac3d

pali:
Thank you! I updated PR

PR #499, opened 2020-11-04
Fix compilation with i586-mingw32msvc-gcc compiler

pali:
Add also i586-mingw32msvc-gcc cross-compile builds to Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,39 @@ jobs:
       env: PROJECT=miniupnpd
     - os: osx
       compiler: gcc
+  include:
+    - os: linux
+      dist: trusty
+      env: PROJECT=miniupnpc
+      compiler: i586-mingw32msvc-gcc
+      addons:
+        apt:
+          packages:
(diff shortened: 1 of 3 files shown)


pali:
Hm... seems that build on Travis was not started...

miniupnp:
By the way, I'm trying to add github workflows/Actions now

pali:
So you have disabled Travis?

miniupnp:
I guess travis is disabled automatically, because I have done nothing on travis configuration

Faless:
This breaks UWP compatibility since `GetBestInterface` is only available for desktop apps:
[GetBestInterfaceEx](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getbestinterfaceex#requirements)
[GetBestInterface](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getbestinterface#requirements)

PR #506, opened 2020-11-22
cmake: update version

RedDwarf69:


miniupnp:
why ?

RedDwarf69:
For no practical difference. Simply because... isn't the current version 2.2?

miniupnp:
nevermind, I was mixing things with the CMake version... :( indeed it should be updated ! thanks

PR #510, opened 2020-11-25
Install cmake config file

RedDwarf69:
I have included all the commits here because there would be merge conflicts otherwise.

CMake usually only builds shared or static, not both. So there isn't really a standard way, but I went for the miniupnpc::miniupnpc alias library which alias the static or shared library based on a MINIUPNPC_USE_STATIC_LIBS variable. The use of XXX_USE_STATIC_LIBS exists on some of the FindXXX modules that come with cmake.


miniupnp:
could you rebase on master ?

miniupnp:
thanks !

Issue #299, opened 2018-05-07
Proxy support in miniupnpd


pali:
Now some ISPs which put end IPv4 users behind Carrier-Grade NATs started supporting PCP.

So if end user with his own IPv4 router (which does NAT for his own local network) wants to use PCP support from ISP, then he needs to run something like PCP proxy on his router.

I would suggest to implement PCP proxy in miniupnpd. When client from internal network ask for port forwarding then miniupnpd would try to request for port forwarding via external interface to upstream ISP network.

This would also allow to connect one miniupnpd router behind another miniupnpd router with working IPv4 address forwarding.

wellloaded:
This is a great idea and I have been looking for something similar for long time. There's a legacy, now unsupported, project that does that as an external package. 
https://github.com/tomaszmrugalski/portproxy

So yes I would absolutely love to have this feature built in miniupnpd too!

pali:
@miniupnp: Do you have any plans for implementing this functionality? It looks like that more people are interested in it and due to higher usage of CG-NATs and DS-Lite, such proxy support would be the only possible way how to use miniupnpd on IPv4 networks without public IPv4 address.

miniupnp:
I need to implement https://tools.ietf.org/html/rfc6970

professor-jonny:
my ISP provided router some how passes upnp requests upstream to my ISP's CG-NAT device.

my ISP's router is tag locked so that the sim card will only work in specific devices where the first 9 digits if the IMEI number must match what is stored in the sim or it will not operate.

From what I understand  UPNP reuests are forwarded upstream as outbount NAT PCP upstream and ports are opened dynamically on the stock router, any therfore any downstream NAT device means that UPNP will fail to operate.

If miniupnpd could be updated to forward upnp requests to an upstream device it would be awesome.

masx200:
https://github.com/phillipberndt/upnprd/

professor-jonny:
> https://github.com/phillipberndt/upnprd/

It would be nice if this could somehow be bundled in with miniupnpd and somehow integrated so requests could be forwarded up stream if it detects an upstream upnp device.

PR #511, opened 2020-11-25
options: ext_ip_reserved_ignore support

ptpt52:
add an option to ignore even if ext_ip is reserved

this make the port forwarding force to work when the router is behind NAT

### DIFF
--- a/miniupnpd/getifaddr.c
+++ b/miniupnpd/getifaddr.c
@@ -25,6 +25,7 @@
 #if defined(USE_GETIFADDRS) || defined(ENABLE_IPV6) || defined(ENABLE_PCP)
 #include <ifaddrs.h>
 #endif
+#include "upnpglobalvars.h"
 
 int
 getifaddr(const char * ifname, char * buf, int len,
@@ -294,6 +295,11 @@ addr_is_reserved(struct in_addr * addr)
 {
 	uint32_t address = ntohl(addr->s_addr);
 	size_t i;
+
+	if(GETFLAG(EXTIPRESERVEDIGNOREMASK)) {
+		syslog(LOG_NOTICE, "private/reserved address checking is ignore");
+		return 0;
+	}
 
 	for (i = 0; i < sizeof(reserved)/sizeof(reserved[0]); ++i) {
 		if ((address >> reserved[i].rmask) == (reserved[i].address >> reserved[i].rmask))
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1225,6 +1225,10 @@ init(int argc, char * * argv, struct runtime_vars * v)
 			case UPNPEXT_IP:
 				use_ext_ip_addr = ary_options[i].value;
 				break;
+			case UPNPEXT_IP_RESERVED_IGNORE:
+				if(strcmp(ary_options[i].value, "yes") == 0)
+					SETFLAG(EXTIPRESERVEDIGNOREMASK);
+				break;
 			case UPNPEXT_PERFORM_STUN:
 				if(strcmp(ary_options[i].value, "yes") == 0)
 					SETFLAG(PERFORMSTUNMASK);
--- a/miniupnpd/miniupnpd.conf
+++ b/miniupnpd/miniupnpd.conf
@@ -9,6 +9,8 @@
 # Setting ext_ip is also useful in double NAT setup, you can declare here
 # the public IP address.
 #ext_ip=
+#ignore even if ext_ip is reserved: default is no
+#ext_ip_reserved_ignore=yes
 # WAN interface must have public IP address. Otherwise it is behind NAT
 # and port forwarding is impossible. In some cases WAN interface can be
 # behind unrestricted NAT 1:1 when all incoming traffic is NAT-ed and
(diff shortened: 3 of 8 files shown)


pali:
Function `addr_is_reserved` checks if address is external public or not and based on it construct reply for `GetExternalIPAddress` UPnP XML function. So this patch breaks `GetExternalIPAddress` function and it returns nonsense/incorrect reply.

Have you tested this change? I'm guess that not because miniupnpc at the first stage checks that `GetExternalIPAddress` returns correct value which results that miniupnpc would be confused or not work.

And even that port forwarding obviously does not work behind NAT (with small exception when full cone NAT 1:1 is used).

I do not know what @miniupnp thinks about it, but this patch does not make sense at all.

pali:
@miniupnp: maybe you should comment this pull request...

miniupnp:
`addr_is_reserved()` behavior should not be changed.

pali:
Any updates in this pull request? Based on my above comments, I do not think that pull request should be merged... so it can be closed?

ptpt52:
so there is no possible to add DNAT rule if router has no public ip? this seems not true in real world

pali:
See my comment, it is working fine if you are behind Full Cone NAT 1:1. Just you need to supply public IP address which is required by UPnP protocol. It can be done either automatically by reading it from wan interface or by discovery via stun code when `ext_perform_stun=yes` is set or manually by command line argument `-o` or by config option `ext_ip`.

pali:
@miniupnp I think you should close this pull request. People now started reporting bugs against this patch (like [this](https://github.com/miniupnp/miniupnp/issues/542)) as openwrt already [expressed](https://github.com/openwrt/packages/issues/15258) that do not maintain this custom patch. It looks like that openwrt has misconfigured iptables/routing (like can be seen in this [patch](https://github.com/openwrt/packages/pull/13951) and links from it) which caused issues in miniupnpd and they are trying to hack them by patch like this one.

pali:
Patch in this pull request is causing more issues (e.g. https://github.com/openwrt/packages/issues/15258 or https://github.com/miniupnp/miniupnp/issues/546 https://github.com/openwrt/packages/pull/14031#issuecomment-851798901) and OpenWRT now decided to disable this option by default due to lot of bug reports https://github.com/openwrt/packages/pull/15759

ptpt52:
close this. 

> `addr_is_reserved()` behavior should not be changed.

sky96111:
> `addr_is_reserved()` behavior should not be changed.

Double NAT are quite common in China. Because optical modem provided by ISP are defaulted with router function and can not be disabled by us. Though we can config DMZ or port forwarding to expose the second router to the public net, miniupnpd can only get Intranet IP and return "addr_is_reserved".

choyri:
Especially for China Mobile, the IP obtained by household PPPoE is all private address. 😞  so sad

sky96111:
> Especially for China Mobile, the IP obtained by household PPPoE is all private address.  so sad

In that case, UPnP still cannot work enen if miniUPnP allow this behaviour. Or you will need to ask ISP to forward your port, but actually it is impossible.
This change only breaks double NAT. For those who can only get a private address from ISP... This breaks nothing.

如果你的光猫也只能从运营商得到一个内网IP，即使这个软件不限制，你的UPnP是不成功的。
那个commit对中国用户有影响，但影响的用户是：光猫拨号能获得外网IP，在光猫上通过端口转发或DMZ把二级路由暴露在了公网上，但是二级路由只能获得光猫分配的内网IP导致miniUPnP拒绝了UPnP。
如果打电话给运营商也不提供公网IP，那换联通电信吧。

miniupnp:
If you only have a Private address, you have not a routable address so using UPnP IGD is pointless.
I hope you still have a routable IPv6 address.

pali:
Obviously if you have only private IPv4 address from CGNAT then UPnP IGD services for port forwarding (like miniupnpd) will never work. To make it work you have to run miniupnpd on that CGNAT machine.

So if you do not have publicly routable IPv4 address then ask for block of routable IPv6 addresses.

But note that IPv4 multi-NAT setup with miniupnpd in cascade (when on every IPv4 CG/NAT router is running miniunpd) is not supported yet by miniupnpd yet, there is open feature request for it https://github.com/miniupnp/miniupnp/issues/299

Neustradamus:
There is a new patch from @ptpt52:
- https://github.com/openwrt/packages/pull/16011#issuecomment-920538698

It replaces the old OpenWrt patch: 301-ext_ip_reserved_ignore.patch.

Tested by @openips!

openips:
if you use miniupnpd as router with more than one isp like me , you can get private ip and public ip at same time. 
  so miniupnpd work with the public ip and not work with the private ip is necessary.
  the new patch let miniupnpd can work in this situation and tested by me.
  great work and thanks a lot

ArcticLampyrid:
> close this.
> 
> > `addr_is_reserved()` behavior should not be changed.

I am against all of this. I am implementing some multi-layer NAT networks, and I hope that port forwarding will be automatically configured through UPnP in my 2nd NAT, even if no public network address is available.

This is a reasonable requirement that will help Tailscale utilize the campus network to establish direct connections (if both parties are located under the same campus's second level NAT)

PR #514, opened 2020-12-12
Fix syntax error in Travis YAML file

pali:
Travis YAML file cannot be currently parsed and therefore Travis was not started, see:
https://travis-ci.org/github/miniupnp/miniupnp/requests


miniupnp:
it fails ;)

pali:
It should be fixed right now after second commit.

pali:
now it passes :-)

PR #516, opened 2020-12-29
Fallback for earlier Apple platforms that lack clock_gettime

past-due:
`clock_gettime` is only available on macOS >= 10.12, iOS >= 10.0 (etc).

Support compiling with a MACOSX_DEPLOYMENT_TARGET < 10.12 by using `__builtin_available` to detect when running on an earlier system, and fall-back to `gettimeofday` when `clock_gettime` is not available.

### DIFF
--- a/miniupnpc/minissdpc.c
+++ b/miniupnpc/minissdpc.c
@@ -463,14 +463,29 @@ static int upnp_gettimeofday(struct timeval * tv)
 	tv->tv_usec = (ts % 1000) * 1000;
 	return 0; /* success */
 #elif defined(CLOCK_MONOTONIC_FAST) || defined(CLOCK_MONOTONIC)
-	struct timespec ts;
-	int ret_code = clock_gettime(UPNP_CLOCKID, &ts);
-	if (ret_code == 0)
-	{
-		tv->tv_sec = ts.tv_sec;
-		tv->tv_usec = ts.tv_nsec / 1000;
+#if defined(__APPLE__)
+#if defined(__clang__)
+	if (__builtin_available(macOS 10.12, iOS 10.0, tvOS 10.0, watchOS 3.0, *)) {
+#else // !defined(__clang__)
+	if (clock_gettime != NULL) {
+#endif // defined(__clang__)
+#endif // defined(__APPLE__)
+		struct timespec ts;
+		int ret_code = clock_gettime(UPNP_CLOCKID, &ts);
+		if (ret_code == 0)
+		{
+			tv->tv_sec = ts.tv_sec;
+			tv->tv_usec = ts.tv_nsec / 1000;
+		}
+		return ret_code;
+#if defined(__APPLE__)
 	}
-	return ret_code;
+	else
+	{
+		// fall-back for earlier Apple platforms
+		return gettimeofday(tv, NULL);
+	}
+#endif // defined(__APPLE__)
 #else
 	return gettimeofday(tv, NULL);
 #endif


miniupnp:
shouldn't `mach_absolute_time()` be used instead ?

past-due:
> shouldn't `mach_absolute_time()` be used instead ?

I hadn't had a chance to dig into where all `upnp_gettimeofday` was used (and how).

Assuming the constraint is not an issue (see below), it's a reasonable option and has been available since macOS 10.0:
> ### mach_absolute_time
> Returns current value of a clock that increments monotonically in tick units (starting at an arbitrary point), **this clock does not increment while the system is asleep**.

A few references that might be helpful: https://developer.apple.com/library/archive/qa/qa1398/_index.html

miniupnpc/minissdpc.c
```diff
@@ -463,14 +463,25 @@ static int upnp_gettimeofday(struct timeval * tv)
 	tv->tv_usec = (ts % 1000) * 1000;
 	return 0; /* success */
 #elif defined(CLOCK_MONOTONIC_FAST) || defined(CLOCK_MONOTONIC)
-	struct timespec ts;
-	int ret_code = clock_gettime(UPNP_CLOCKID, &ts);
-	if (ret_code == 0)
+#if defined(__APPLE__)
```
miniupnp, line 466: when `clock_gettime()` is not available, `CLOCK_MONOTONIC_FAST` and `CLOCK_MONOTONIC` are not defined, right ? so it is already falling back to `gettimeofday()`, right ?

past-due: @miniupnp:
AFAIK, When compiling with a modern macOS SDK:
- `CLOCK_MONOTONIC` and `clock_gettime` are both defined, regardless of `MACOSX_DEPLOYMENT_TARGET`
- They are guarded with availability notations, signifying that if you set (for example) a `MACOSX_DEPLOYMENT_TARGET` < 10.12, they will not *work* on prior platforms (and will be weak-linked).
- Instead of doing things like manually checking if `clock_gettime` is `NULL` at runtime (which will not silence the warnings), the recommended solution is using `__builtin_available`, which handles only executing the branch if the code is running on supported versions of Apple platforms.

So, no, if compiling on a modern macOS SDK: `clock_gettime` and `CLOCK_MONOTONIC` are always defined in the headers, regardless of whether they will actually be available at run-time. Hence the warning, and - while I haven't tested this specific case - likely a crash when running on earlier versions of Apple platforms (without a change like the above).

miniupnp: is  `__builtin_available() ` when compiling on every Apple platform / compiler ? I'm afraid it will prevent buildin with gcc.

I would prefer #undef CLOCK_MONOTONIC_FAST and CLOCK_MONOTONIC if MACOSX_DEPLOYMENT_TARGET < 10.12

past-due: @miniupnp:

`__builtin_available` is indeed Clang-specific, but works for every Apple platform (assuming Clang / Xcode). It also does some "fancy" things depending on what the minimum deployment target is, to avoid any penalty unless necessary (IIRC).

> I would prefer #undef CLOCK_MONOTONIC_FAST and CLOCK_MONOTONIC if MACOSX_DEPLOYMENT_TARGET < 10.12

AFAIK, `MACOSX_DEPLOYMENT_TARGET` isn't passed as a define to source code. It's used, as an environment variable, to default various flags sent to the compiler / linker. (Those flags can also be specified directly - ex `-mmacosx-version-min=10.12`.)

However if you'd like to make this smart for clang, and "working" for gcc, it could be modified to something like:
```
#if defined(__APPLE__)
#if defined(__clang__)
	if (__builtin_available(macOS 10.12, iOS 10.0, tvOS 10.0, watchOS 3.0, *)) {
#else
	if (clock_gettime != NULL) {
#endif // defined(__clang__)
```

My understanding is this will do "the proper thing" for clang / Xcode (with all the advantages that come with it, including smart behavior depending on deployment target and suppressing the warning / error), and will avoid calling a weak-linked `clock_gettime` on other compilers (which _should_ be set to NULL, and presumably will lack the warning / error).

past-due: I have updated the PR with my suggested strategy above to maintain non-clang compatibility.

past-due:
@miniupnp: This updated PR now handles the non-clang compatibility case you described. (Please see my comments above for my rationale.) Were there additional changes you wanted to get this merged?

PR #518, opened 2020-12-31
miniupnpc: Fix usage of IP_MULTICAST_IF with struct ip_mreqn

pali:
When struct ip_mreqn is passed to IP_MULTICAST_IF setsockopt option it is
always required to set also ipv4 source address. Otherwise Linux kernel
will choose default system multicast ipv4 address which does not have to
belong to chosen interface specified in struct ip_mreqn.

Therefore on system with more multicast interfaces and more ipv4 addresses,
it may happen that interface chosen by upnpc -m option would use ipv4
address which does not belong to this interface.

This change is fixing above issue and ensure that if interface is chosen by
upnpc -m option then source address which belongs to this interface would
be used.

Without this change upnpc -m eth1 can send multicast traffic over interface
eth1 but with source ipv4 address of interface eth0, which obviously would
be rejected by upnp gateway.

### DIFF
--- a/miniupnpc/minissdpc.c
+++ b/miniupnpc/minissdpc.c
@@ -697,6 +697,13 @@ ssdpDiscoverDevices(const char * const deviceTypes[],
 			 * MS Windows Vista and MS Windows Server 2008.
 			 * http://msdn.microsoft.com/en-us/library/bb408409%28v=vs.85%29.aspx */
 			unsigned int ifindex = if_nametoindex(multicastif); /* eth0, etc. */
+			if(ifindex == 0)
+			{
+				if(error)
+					*error = MINISSDPC_INVALID_INPUT;
+				fprintf(stderr, "Invalid multicast interface name %s\n", multicastif);
+				goto error;
+			}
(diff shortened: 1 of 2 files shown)


miniupnpc/minissdpc.c
```diff
@@ -728,29 +735,40 @@ ssdpDiscoverDevices(const char * const deviceTypes[],
...
 				ifr.ifr_name[IFNAMSIZ-1] = '\0';
 				if(ioctl(sudp, SIOCGIFADDR, &ifr, &ifrlen) < 0)
 				{
 					PRINT_SOCKET_ERROR("ioctl(...SIOCGIFADDR...)");
+					goto error;
 				}
 				mc_if.s_addr = ((struct sockaddr_in *)&ifr.ifr_addr)->sin_addr.s_addr;
+#ifdef HAS_IP_MREQN
+				/* was not an ip address, try with an interface name */
+				struct ip_mreqn reqn;	/* only defined with -D_BSD_SOURCE or -D_GNU_SOURCE */
```
miniupnp, line 751: there is an issue with moving the ` struct ip_mreqn reqn;` declaration in the middle of the block, it won't build with C90 compilers

pali: I fixed it now.

PR #520, opened 2021-01-08
submit_to_minissdpd: module, hardening, RC plus Message 

sanderjo:


minissdpd/submit_to_minissdpd.py
```diff
@@ -1,32 +1,40 @@
...
-    encodedlen = b''
+        return b"\x00"
+    encodedlen = b""
     while l > 0:
         c = l & 0x7F
         l = l >> 7
         if l > 0:
             c = c + 128
-        encodedlen = c.to_bytes(1, 'little') + encodedlen
+        encodedlen = c.to_bytes(1, "little") + encodedlen
```
miniupnp, line 27: why do you convert single quotes to double quotes ?

sanderjo: Ah, good point. Answer: code formatter `black `did that. 
PEP8 does **not** specify what it should be: https://www.python.org/dev/peps/pep-0008/#:~:text=In%20Python%2C%20single%2Dquoted%20strings,avoid%20backslashes%20in%20the%20string. So python doesn't care.

A long discussion on black's behaviour resulted in https://github.com/psf/black/issues/118#issuecomment-393298377 , so with option -S I can instruct black to ignore single / double quotes.

So: if you want the single quotes back, I can do that (and still use black). Let me know.

miniupnp:
I cherry-picked and modified.

PR #526, opened 2021-01-27
miniupnpc: Fix compile warnings

pali:


pali:
Thank you for review & merge!

Now when in git master branch are compile fixes for Visual Studio, fixes for `upnpc -m` and updated hints/error messages from stun checks, what about releasing a new minor version?

PR #530, opened 2021-02-24
Fix incorrect addition of jump rule to PREROUTING mangle chain

albertmichaelj:
In the current iptables init script, the mangle table has a rule added to jump to the MINIUPNPD chain in the mangle table added to PREROUTING. However, if you look at `iptables_removeall.sh`, it tries to remove the rule from the `FORWARD` chain instead (see the line [here](https://github.com/miniupnp/miniupnp/blob/bfc5eab6be38d693d0fa8bfecc467021c0c85c1f/miniupnpd/netfilter/iptables_removeall.sh#L21). This changes the init script to add the jump rule to the FORWARD chain instead of the PREROUTING chain.

When this is not done, the mangle table is not properly reset on calling the `iptables_removeall.sh` script. Specifically the rule jumping to MINIUPNPD and the MINIUPNPD chain itself remain (though rules are correctly flushed). Additionally, there is an error message in the logs as follows:

```
REDACTED iptables_removeall.sh[1850]: iptables: No chain/target/match by that name.
REDACTED iptables_removeall.sh[1851]: iptables: Too many links.
```

To be frank, I am not certain that the intended chain to add the jump rule to is the FORWARD chain. The mistake might be in `iptables_removeall.sh` in trying to remove the rule from the `FORWARD` chain. However, to the best I can understand, the mangle table is used to adjust the DSCP of incoming packets (though to be clear, I don't understand when and why this is done in the UPnP protocol), and by looking at other examples online, it seems like the FORWARD chain on the mangle table is the most common place to adjust DSCP values. Additionally, the git blame for the line in `iptables_removeall.sh` shows that the FORWARD chain was used first (`iptables_removeall.sh` was edited nine years ago to add the removal rule, while `iptables_init.sh` was added two years ago). All of this leads me to believe that the correct chain is the FORWARD chain.

### DIFF
--- a/miniupnpd/netfilter/iptables_init.sh
+++ b/miniupnpd/netfilter/iptables_init.sh
@@ -32,11 +32,11 @@ if [ "$MDIRTY" = "${CHAIN}Chain" ]; then
 	echo "Mangle table dirty; Cleaning..."
 elif [ "$MDIRTY" = "Chain" ]; then
 	echo "Dirty Mangle chain but no reference..? Fixing..."
-	$IPTABLES -t mangle $ADDCMD PREROUTING -i $EXTIF -j $CHAIN
+	$IPTABLES -t mangle $ADDCMD FORWARD -i $EXTIF -j $CHAIN
 else
 	echo "Mangle table clean..initializing..."
 	$IPTABLES -t mangle -N $CHAIN
-	$IPTABLES -t mangle $ADDCMD PREROUTING -i $EXTIF -j $CHAIN
+	$IPTABLES -t mangle $ADDCMD FORWARD -i $EXTIF -j $CHAIN
 fi
 if [ "$CLEAN" = "yes" ]; then
 	$IPTABLES -t mangle -F $CHAIN


miniupnp:
DSCP is used for PCP protocol, not UPnP.
support was added in 9e1ffd5cd9836053cfd83d95014c729ad0e36872

```
diff --git a/miniupnpd/netfilter/iptables_init.sh b/miniupnpd/netfilter/iptables_init.sh
index d3510dc..0b064c9 100755
--- a/miniupnpd/netfilter/iptables_init.sh
+++ b/miniupnpd/netfilter/iptables_init.sh
@@ -13,8 +13,15 @@ $IPTABLES -t nat -N MINIUPNPD
 #$IPTABLES -t nat -A PREROUTING -d $EXTIP -i $EXTIF -j MINIUPNPD
 $IPTABLES -t nat -A PREROUTING -i $EXTIF -j MINIUPNPD
 
+#adding the MINIUPNPD chain for mangle
+$IPTABLES -t mangle -N MINIUPNPD
+$IPTABLES -t mangle -A PREROUTING -i $EXTIF -j MINIUPNPD
+
 #adding the MINIUPNPD chain for filter
 $IPTABLES -t filter -N MINIUPNPD
 #adding the rule to MINIUPNPD
 $IPTABLES -t filter -A FORWARD -i $EXTIF ! -o $EXTIF -j MINIUPNPD
 
+#adding the MINIUPNPD chain for nat
+$IPTABLES -t nat -N MINIUPNPD-PCP-PEER
+$IPTABLES -t nat -A POSTROUTING -o $EXTIF -j MINIUPNPD-PCP-PEER
```

miniupnp:
somehow, the error was introduced in `iptables_removeall.sh` in 82ec7bc3df0eab362163545f5e61eb97a84652cd

miniupnp:
using PREROUTING was intended. Unless there is a good reason to change, I keep it that was ;)

I have fixed `iptables_removeall.sh` with https://github.com/miniupnp/miniupnp/commit/9239cf28c1ddf20e46974f464e901b8ad0385d5e

albertmichaelj:
All this looks good to me! Like I said, I didn't know on which side the error was, so I'm happy to keep it at PREROUTING. Thanks for your work on this very useful piece of software!

PR #533, opened 2021-03-18
miniupnpd: Add option to match rules with regex

yangfl:
Some reports that a certain app is abusing UPnP for exploiting upload
bandwidth. This commit adds support to restrict UPnP rules to a regex.
By matching requester's description string against rule's regex, this
will make some obstacles for that app.

### DIFF
--- a/miniupnpd/miniupnpd.conf
+++ b/miniupnpd/miniupnpd.conf
@@ -168,7 +168,7 @@ uuid=00000000-0000-0000-0000-000000000000
 #force_igd_desc_v1=no
 
 # UPnP permission rules
-# (allow|deny) (external port range) IP/mask (internal port range)
+# (allow|deny) (external port range) IP/mask (internal port range) (optional regex filter)
 # A port range is <min port>-<max port> or <port> if there is only
 # one port in the range.
 # IP/mask format must be nnn.nnn.nnn.nnn/nn
@@ -180,6 +180,8 @@ uuid=00000000-0000-0000-0000-000000000000
 # also consider implementing network-specific restrictions
 # CAUTION: failure to enforce any rules may permit insecure requests to be made!
 allow 1024-65535 192.168.0.0/24 1024-65535
+# disallow requests whose description string matches the given regex
+# deny 1024-65535 192.168.1.0/24 1024-65535 "My evil app ver [[:digit:]]*"
 allow 1024-65535 192.168.1.0/24 1024-65535
 allow 1024-65535 192.168.0.0/23 22
 allow 12345 192.168.7.113/32 54321
--- a/miniupnpd/upnppermissions.c
+++ b/miniupnpd/upnppermissions.c
@@ -13,8 +13,151 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <unistd.h>
+
+#ifdef ENABLE_REGEX
+#include <regex.h>
+#endif
+
 #include "config.h"
+#include "macros.h"
 #include "upnppermissions.h"
+
+static int
+isodigit(char c)
+{
+	return '0' <= c && c >= '7';
+}
+
+static char
+hex2chr(char c)
+{
+	if(c >= 'a')
+		return c - 'a';
+	if(c >= 'A')
+		return c - 'A';
+	return c - '0';
+}
+
+static char
+unescape_char(const char * s, int * seqlen)
+{
+	char c;
+	int len;
+
+	if(s[0] != '\\')
+	{
+		c = s[0];
+		len = 1;
+	}
+	else
+	{
+		s++;
+		c = s[0];
+		len = 2;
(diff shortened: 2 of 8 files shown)


miniupnp:
Interesting.
@yangfl  can you link to the reference of the problem ?

yangfl:
https://t.me/LeanSaidWTF/198

miniupnpd/upnppermissions.c
```diff
@@ -16,6 +17,143 @@
 #include "config.h"
 #include "upnppermissions.h"
 
+static int
+ishexdigit(char c)
+{
+	return ('0' <= c && c >= '9') || ('A' <= c && c >= 'F') ||
+	       ('a' <= c && c >= 'f');
+}
```
pali, line 25: Duplicate of standard C89/C99 function `isxdigit` from `ctype.h`

miniupnpd/miniupnpd.conf
```diff
@@ -176,7 +176,7 @@ uuid=00000000-0000-0000-0000-000000000000
 # also consider implementing network-specific restrictions
 # CAUTION: failure to enforce any rules may permit insecure requests to be made!
```
pali, line 177: I would suggest to add here some documentation. Including what special string `NAT-PMP` means.

miniupnpd/upnppermissions.c
```diff
@@ -16,6 +17,143 @@
...
+	case 'n':  c = '\n'; break;
+	case 'r':  c = '\r'; break;
+	case 't':  c = '\t'; break;
+	case 'v':  c = '\v'; break;
+	case '\\': c = '\\'; break;
+	case '\'': c = '\''; break;
+	case '"':  c = '"';  break;
+	case '?':  c = '?';  break;
+	case 'x':
+		if(ishexdigit(s[1]) && ishexdigit(s[2]))
```
pali, line 73: For me it looks like that this code can cause memory corruption. There is missing check that there exist some character after `'x'`.

miniupnpd/upnppermissions.c
```diff
@@ -16,6 +17,143 @@
...
+		{
+			c = hex2chr(s[1]) * 0x10 + hex2chr(s[2]);
+			len = 4;
+		}
+		break;
+	default:
+		if(isoctdigit(s[1]) && isoctdigit(s[2]) && isoctdigit(s[3]))
```
pali, line 80: And same here. As there is no check that some character continue after `'\\'`.

miniupnpd/upnppermissions.c
yangfl: nope, since `isxdigit(0) == 0`.

miniupnpd/miniupnpd.conf
yangfl: I'd like to fulfill the documentation after this pr get merged.

miniupnpd/upnppermissions.c
pali: Now I see. But it is very non-intuitive work with c strings here which I think after future modifications can lead to CVE.

It really is not intuitive that you cannot call `unescape_char` on empty string or that `isxdigit` is used for implicit checking of end-of-string.

yangfl: well `isdigit` is clearly stated that it can check EOF... but a redundant check can be added if that is perferred by the coding style of this repo (and `unescape_char` can be called on empty string, in fact)

pali: Well, @miniupnp is maintainer here, so wait for his opinion.

> (and unescape_char can be called on empty string, in fact)

Ok, now I figured out that you are right here too! But my first impression after looking at the code was that this can cause out-of-bounds read.

miniupnpd/upnppermissions.c
```diff
@@ -172,15 +305,61 @@ read_permission_line(struct upnpperm * perm,
 	{
 		return -1;
 	}
+
+	/* fifth token: (optional) regex */
+	p = get_next_token(p, &perm->re, 1);
```
miniupnp, line 310: why is get_next_token() needed ?
Isn't taking the end of the line OK ?

yangfl: so that forward compatibility with more option tokens is possible.

miniupnp:
I'm sorry I answer with such delay.

I think I will include this code but with an option to disable it at compile time.
There may be performance issues when running regex so people that want to disable that code must be able to do it.

yangfl:
Oh, so do you want me to rework this pr and add a compile option? ;)

yangfl:
Thank you! This should not impact performance of existing configs since regex matching is not called if no regex filter defined. Nevertheless I've add such an option for it.

miniupnp:
see 2ff8cb17da6485c3bdd2503e7e81931ddd68b6b6

PR #537, opened 2021-03-28
IGD v2.0: Unknown ExternalIPAddress is represented as empty string

pali:
IGD v2.0 specification for WANIPConnection:2 says:
    
When the external IP address could not be retrieved by the gateway (for
example, because the interface is down or because there was a failure in
the last connection setup attempt), then the ExternalIPAddress MUST be
equal to the empty string.


miniupnp:
@pali thanks !

PR #553, opened 2021-07-15
miniupnpd: don't check for glibc version with musl

ncopa:
Test that ldd is from GLIBC before exctracting the GLIBC_VERSION. This
is not needed with musl libc.

### DIFF
--- a/miniupnpd/Makefile.linux
+++ b/miniupnpd/Makefile.linux
@@ -95,6 +95,7 @@ $(info please install uuid-dev package / libuuid)
 endif # ($(TEST),1)
 endif # ($(TARGET_OPENWRT,)
 
+ifneq ($(shell ldd --version | grep GLIBC),)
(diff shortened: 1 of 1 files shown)


miniupnpd/Makefile.linux
```diff
@@ -95,13 +95,15 @@ $(info please install uuid-dev package / libuuid)
 endif # ($(TEST),1)
 endif # ($(TARGET_OPENWRT,)
 
+ifneq ($(shell --ldd --version | grep GLIBC),)
```
miniupnp, line 98: it should be `$(shell ldd --version | grep GLIBC)`

miniupnp: as it is :
```
/bin/sh: 0: Illegal option --
```

ncopa: doh... ofc..

PR #557, opened 2021-08-06
miniupnpd: Disable port forwarding when upstream interface is down

pali:
Obviously port forwarding cannot work when upstream interface is down. So
correctly report status code for port forwarding requests to clients in
this case.


miniupnp:
I think it is missing some code to enable again when network is up again

pali:
Reenabling is already implemented in this block:
https://github.com/miniupnp/miniupnp/blob/7ee554d31b47a7227ab85aa919792597ce78c81e/miniupnpd/miniupnpd.c#L2524-L2535

PR #559, opened 2021-08-13
Include dir

### LINKED ISSUE #540: VERSION file causes problems when compiled with C++20
In C++20, `<version>` is a header the compiler looks for. When miniupnp is used as a submodule, the submodule's `miniupnpc` directory is often set as a header location. On case insensitive file systems (macOS default for instance), this causes compilation errors since the VERSION file is incorrectly picked up as the `<version>` header.

The ideal IMO would be if miniupnpc didn't mix source and non-source files (maybe a `src` or at least an `include` directory), while renaming `VERSION` is another solution (the least breaking change). It can also be solved various ways for each user, but would be nice if it worked out of the box :) Thanks for considering.

miniupnp:
fixes #540


miniupnpc/CMakeLists.txt
```diff
@@ -56,40 +56,24 @@ configure_file (${CMAKE_CURRENT_SOURCE_DIR}/miniupnpcstrings.h.cmake ${CMAKE_CUR
 target_include_directories(miniupnpc-private INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
 
 set (MINIUPNPC_SOURCES
-  igd_desc_parse.c
-  miniupnpc.c
-  minixml.c
-  minisoap.c
-  minissdpc.c
```
scribam, line 63: Because `minissdpc.c` has been removed here, I have the following error on Windows (MSVC) with CMake:
```
libminiupnpc.lib(miniupnpc.c.obj) : error LNK2019: unresolved external symbol ssdpDiscoverDevices referenced in function upnpDiscover
```

miniupnp: that should be fixed in 2017ae036f67f1a47b49191b72a6ac9b262fda3e

scribam: I confirmed, it works now, thank you.

Issue #18, opened 2012-08-31
IPv6 "pinholes" are not added to lease_file


atomt:
Title says it all ;-)

When adding firewallopening using WANIPv6FirewallControl the rules does not get persisted to leases file.
Only IPv4 mappings gets saved.

noloader:
@miniupnp,

Please forgive my ignorance since I don't do much work on Windows anymore... Do modern Microsoft products, like Windows 8 and Windows 10, honor the UPnP commands like punching a hole in the Windows firewall?

I read Microsoft's [Writing Secure Code for Windows Vista](http://www.amazon.com/dp/0735623937) guide twice. There's another interface (Win32 API) that's supposed to be used. But the book is older, and I'm wondering if/how things have changed.

miniupnp:
I'm sorry this is offtopic.
And I just don't know.

treysis:
Unfortunately that seems like a lot of work. Not only will you have to write to the file (should be pretty easy), but you would also have to adjust the parsing for IPv4 so that it doesn't get mixed up. And probably in a way to not break too much backwards compatibility.
Like I said, adding seems easy, but I don't feel ready for the parsing part :(

chrysn:
There seems to be two purposes for the lease files: reloading from a lease file (to ensure old leases stay up even after a config change), and introspection from outside (eg. in openwrt's luci-app-upnp, which displays a refreshing table in the router status page).

I can't easily judge the relevance of IPv6 addresses for the former (as it appears to be primarily for the benefit of UPnP, given that PCP clients would use announce-based rapid recovery). Might it be sufficient to export the data in a non-backparseable way just for the benefit of the latter purpose?

Neustradamus:
Any news on it?

svenauhagen:
I have some code for this issue. I will clean it up and send a PR the next days.

Neustradamus:
@svenauhagen: A nice contribution, thanks!

PR #561, opened 2021-08-17
IPv6 pinholes lease file

svenauhagen:
This patch adds a lease file for IPv6 pinholes.
The leases are maintained and readded when miniupnpd restarts.
Currently all IPv6 leases are lost on restart.

There is an open request for #18 

Please let me know if there should be any changes.

Best
Sven

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1441,6 +1441,11 @@ init(int argc, char * * argv, struct runtime_vars * v)
 			case UPNPLEASEFILE:
 				lease_file = ary_options[i].value;
 				break;
+#ifdef ENABLE_UPNPPINHOLE
+			case UPNPLEASEFILE6:
+				lease_file6 = ary_options[i].value;
+				break;
+#endif	/* ENABLE_UPNPPINHOLE */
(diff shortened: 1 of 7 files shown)


miniupnp:
I guess it is missing a few `#ifdef ENABLE_IPV6`

miniupnp:
> I guess it is missing a few `#ifdef ENABLE_IPV6`

or `#ifdef ENABLE_UPNPPINHOLE`

svenauhagen:
Hi, I used #ifdef ENABLE_UPNPPINHOLE in most of the places.
The entire file upnppinhole.c is wrapped in ENABLE_UPNPPINHOLE.
I only left it out in the globavars and options files.
Do you think it it better to add it there as well?

miniupnp:
> Hi, I used #ifdef ENABLE_UPNPPINHOLE in most of the places.
> The entire file upnppinhole.c is wrapped in ENABLE_UPNPPINHOLE.
> I only left it out in the globavars and options files.
> Do you think it it better to add it there as well?

yes.

svenauhagen:
I updated the commit and wrapped everything in a #ifdef ENABLE_UPNPPINHOLE

miniupnp (approved):

Issue #422, opened 2020-03-03
nftables: 'miniupnpd' chain gets overruled by default drop policy on base chain


NicholasFahey:
Trying to get miniupnpd working with nftables and I am _most_ of the way there, but I believe there is a still a fundamental issue with how the nftables support is implemented in miniupnpd.

Kernel: Linux 5.3.6
miniupnpd version: 2.1.20191006

My base ruleset is this:
```
table ip filter {
        chain input {
		type filter hook input priority 0; policy drop;
                # Accept some LAN traffic and established WAN traffic here; drop everything else
               ...
	}
	chain forward {
		type filter hook forward priority 0; policy drop;
                # Accept LAN traffic and established WAN traffic here; drop everything else
                ...
        }
}

table ip nat {
	chain prerouting {
		type nat hook prerouting priority -100; policy accept;
	}

	chain postrouting {
		type nat hook postrouting priority 100; policy accept;
		oifname "eth2" masquerade;
	}
}
```
After starting miniupnpd and having it add a port forwarding rule I end up with this in the `miniupnpd` tables:
```
...
table inet miniupnpd {
	chain forward {
		type filter hook forward priority -25; policy accept;
                iif "eth2" @th,16,16 40005 @nh,128,32 3232260974 @nh,72,8 6 accept
	}
}
table ip miniupnpd {
	chain prerouting {
		type nat hook prerouting priority -100; policy accept;
                iif "eth2" tcp dport 40005 dnat to 192.168.99.110:40005
	}

	chain postrouting {
		type nat hook postrouting priority 100; policy accept;
	}
}
table ip6 miniupnpd {
	chain prerouting {
		type nat hook prerouting priority -100; policy accept;
	}

	chain postrouting {
		type nat hook postrouting priority 100; policy accept;
	}
}
```
Those rules look correct but I'm still not able to pass traffic on that port. I believe the issue is that the traffic has to be accepted by _both_ the miniupnpd forward chain _and_ my filter table forward chain. If I add in packet counters to the 2 miniupnpd rules I can see that the packets hit those rules but still end up being dropped despite the higher priority on the miniupnpd chain.

To be honest, nftables docs is a bit unclear in this regard but I think this line from [these docs](https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains) lays it out pretty clearly: 

[...]

PR #562, opened 2021-08-18
NFTables use inet nat chain

svenauhagen:
NFTables supports inet in the nat chain as well.
Use it instead of IPv4 chain so it is consistent with the filter chain.
This also removes the need for two separate tables to manage.

This is the linked to #422 and the first commit to fix it.

Signed-off-by: Sven Auhagen <sven.auhagen@voleatech.de>

### DIFF
--- a/miniupnpd/netfilter_nft/nftnlrdr.c
+++ b/miniupnpd/netfilter_nft/nftnlrdr.c
@@ -84,29 +84,12 @@ init_redirect(void)
 						  nft_forward_chain, FILTER_CHAIN_TYPE, NF_INET_FORWARD, NF_IP_PRI_FILTER - 25);
 	}
 
-	/* 'ip' family */
 	if (result == 0) {
-		result = table_op(NFT_MSG_NEWTABLE, NFPROTO_IPV4, nft_table);
-	}
-	if (result == 0) {
-		result = chain_op(NFT_MSG_NEWCHAIN, NFPROTO_IPV4, nft_table,
+		result = chain_op(NFT_MSG_NEWCHAIN, NFPROTO_INET, nft_table,
 						  nft_prerouting_chain, NAT_CHAIN_TYPE, NF_INET_PRE_ROUTING, NF_IP_PRI_NAT_DST);
 	}
 	if (result == 0) {
-		result = chain_op(NFT_MSG_NEWCHAIN, NFPROTO_IPV4, nft_table,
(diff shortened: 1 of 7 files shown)


miniupnp:
@svenauhagen thank you for the work. I supposed it is well tested.

I guess I should release another 2.2.X version with minor fixes before merging this.

by the way, signaling this is linked to #422

svenauhagen:
@miniupnp I miniupnpd with that patch for the past year and I tested it again the last days on the master branch to make sure it works.

I have two more PRs coming based on this one to fix #422 .
I split them up so it is easier to see what has changed.

miniupnp:
@svenauhagen so I merge this one and you have more coming ?

svenauhagen:
@miniupnp yes, I have two more coming:

https://github.com/svenauhagen/miniupnp/commits/feature/nftablesnatname
and
https://github.com/svenauhagen/miniupnp/commits/feature/nftablesinit

I can create the next PR when you merged this one.

sjnewbury:
I've been unable to upgrade since this commit, presumably due to my firewall having separate rules for ipv4 and ipv6.  I'm using a combination of firewalld + miniupnpd + libvirt + custom nft rules.  It all works perfectly as long as I don't upgrade miniupnpd.  Would it be possible to make it configurable whether to use inet or separate v4/v6 tables?  Or should it be working, and there's something wrong in my configuration?

svenauhagen:
Hi, no it is not possible to use seperate v4/v6 tables with miniupnpd anymore. The backend code was more complicated and we had to distinguish between the two while nftables moves more and more in the direction that you can simply use inet tables and chains which can hold ip4 or ip6 regardless of the actual data in it.
The easiest for you is to change your firewalld to use the inet tables instead of separating ip4 and ip6.

sjnewbury:
Yeah, I understand the decision has been made.  That won't be possible for my use case though,  I guess I'll be sticking with a fork of the version before the feature merge.  I'll try to merge the non-conflicting commits moving forward.

I'm pretty sure nftables isn't dropping support for handling the separate protocol families.  My understanding was "inet" was just introduced as a convenience table for simple firewalls.  It is notably less secure since you'll be opening holes in the firewall where the intended server may not be listening or could be improperly configured.

mooninite:
@sjnewbury the problem is that the miniupnpd default rules have the forward table policy set to 'drop' and this overrides anything else even if you have a chain with a higher priority.

I also run separate 'ip' and 'ip6' family firewall rules and I "fixed" miniupnpd by changing nft_init.sh from "policy drop" to "policy accept". I am not proposing this as a permanent fix at this time, but I would be interested to know if it helps you.

PR #568, opened 2021-09-29
Makefile: fix install rules to account for $(BUILD)

mgorny:
Fix missing references to $(BUILD) in the install rules, as well
as incorrect shared library symlink name.  Otherwise, the 'install'
phase fails either being unable to find files:

    make: *** No rule to make target 'miniupnpc.pc', needed by 'install'.  Stop.

or trying to create the symlink in the wrong subdirectory:

    ln: failed to create symbolic link '/tmp/z/usr/lib/build/libminiupnpc.so': No such file or directory


miniupnp:
@mgorny thank you !

PR #569, opened 2021-09-30
miniupnpc/CMakeLists.txt: fix install for headers

heitbaum:
Install target fails as include files have moved

```
"build/miniupnpc-2.2.3/miniupnpc.h":
  No such file or directory.
FAILED: CMakeFiles/install.util 
cd build/miniupnpc-2.2.3/.x86_64-libreelec-linux-gnu && toolchain/bin/cmake -P cmake_install.cmake
ninja: build stopped: subcommand failed.
```


miniupnp:
thank you

PR #581, opened 2021-11-17
NFTables make tables name configurable

svenauhagen:
Right now the table names are hardcoded and do not integrate with an overall
firewall strategy.
NFTables has restrictions on how packets are evaluated against chains.
For example if multiple forward chains are evaluated with different prioity,
all packets that pass the first one will be reevaluated again in the second chain.
To have an overall firewall concept with miniupnpd it is necessary to use existing
tables and hence to configure them in miniupnpd.

This is the second of three PRs and it is build upon #562.

Signed-off-by: Sven Auhagen <sven.auhagen@voleatech.de>

### DIFF
--- a/miniupnpd/commonrdr.h
+++ b/miniupnpd/commonrdr.h
@@ -77,6 +77,7 @@ update_portmapping_desc_timestamp(const char * ifname,
 
 typedef enum {
 	RDR_TABLE_NAME,
+	RDR_NAT_TABLE_NAME,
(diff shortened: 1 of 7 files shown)


miniupnp:
I have to make sure documentation is up to date

miniupnp:
I rebased.
see 0b3f3e40295288aadb6bcd375deee4ab840daf1d and d4849fa08e38c3f24acd4f4335ac1f95c408aa9a

svenauhagen:
Thank you, next PR is coming as well

PR #586, opened 2022-01-01
A few nftables fixes

f1sherman:
Hello! I was looking through the nftables-related code to try to understand how it works and decided to make a few small improvements. Would love your feedback!

1. Spelling fix: routeing --> routing
2. Fix typo where we're incorrectly checking for `$TABLE` instead of `$NAT_TABLE` in `nft_removeall.sh`
3. Split `NAT` and `TABLE` everywhere for consistency

### DIFF
--- a/miniupnpd/miniupnpd.conf
+++ b/miniupnpd/miniupnpd.conf
@@ -75,7 +75,7 @@
 
 # table names for netfilter nft. Default is "filter" for both
 #upnp_table_name=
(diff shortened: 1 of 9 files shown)


miniupnp:
@svenauhagen is that OK ? :)

miniupnp (approved):
everything looks great

svenauhagen:
looks good to me :) thanks

PR #592, opened 2022-01-24
Refine License

stokito:
The same as for https://github.com/miniupnp/libnatpmp/pull/23
But I also renamed LICENSE to LICENSE.txt so Windows users can easily open it.


miniupnp:
I don't care much about windows Users, and I must say there is no windows users for miniupnpd or minissdpd

PR #604, opened 2022-04-01
cmake: Fix static library name in mingw

Biswa96:
This corrects the static library name to libminiupnpc.a.
Otherwise 'lib' is prepended twice, like liblibminiupnpc.a.


miniupnp (approved):
I hope it works well :)

PR #607, opened 2022-04-18
Fix uninitialized read in update_portmapping

Teemperor:
None of the called functions seemed to use rhost aside from `add_filter_rule2`
which tries to read the uninitialized data in the buffer via:

```
if (rhost != NULL && strcmp(rhost, "") != 0 && strcmp(rhost, "*") != 0) {
  rhost_addr = inet_addr(rhost);
}
```

This randomly tries to parse the uninitialized memory in `rhost`.


miniupnp:
thanks for the report. I think it is better to properly initialize iaddr_str and rhost, so it is possible use it someday.

PR #633, opened 2022-10-21
miniupnpc: update cmake project version

scribam:


miniupnp:
:( I have to fix my script that updates versions :(

Issue #616, opened 2022-05-29
PyPI artifacts


altendky:
I have a couple questions and requests regarding the artifacts on PyPI.  I would expect to put in my own time to achieve these goals, but I should start by understanding your setup and if there are things you would like to see change.

1) Was there a specific need for eggs at this point?
2) 2.2.3 is missing Windows wheels for Python 3.8.  https://pypi.org/project/miniupnpc/2.2.3/#files
3) It would be nice to have Python 3.10 wheels added.
4) It would be nice to have wheels available for all platforms.  Linux, macOS, and Windows.

This should all be doable on just GitHub Actions.  I see configuration files for several CI platforms right now.  How do you see your present CI setup and do you have plans for its future?

altendky:
I should clarify that I am looking to discuss the Python client.  It looks like perhaps @mrx23dot has interest and existing involvement in this as well.

miniupnp:
If generating python wheels with github actions is doable, just do it !
#573 is stuck somehow...

kelsos:
It should be doable to build wheels on GitHub Actions using https://cibuildwheel.readthedocs.io/en/stable/

altendky:
We have interest in the arm stuff as well, but we only have private runners presently.  I think GitHub is working on the macOS ARM runners though, at least.  If we get to this then we would be shooting for the best platform coverage we can get.  And yes, I would like the package to be all inclusive as is the expectation for Python packages (outside of specific reasonably expected system dependencies, like manylinux formalizes).  Though, I don't know if there are special reasons with this project to not have that.

kelsos:
Yes they don't have arm builders at the moment unfortunately, which makes my life really hard.

It is however possible to cross-build arm64 wheels on the x86_64 macOS runners, but it can become really complicated depending the rest of your dependencies.

miniupnp:
3.8 and 3.10 added to https://pypi.org/project/miniupnpc/2.2.3/#files

miniupnp:
anyone able to automatically build macOS wheels ?

kelsos:
I have a script to build a universal2 wheel for packaging with PyInstaller.

I am currently traveling, so I don't have access to the mac, but maybe if I have enough time next week, I can try it and create a PR. If the individual and universal2 wheels are published on PyPI, it will also help us simplify the build script.

altendky:
[...]

PR #640, opened 2023-01-10
build miniupnpc wheels in ci

altendky:
This adds wheel building CI in GitHub Actions for all operating systems and platforms except for Windows ARM. 
 This was requested in https://github.com/miniupnp/miniupnp/issues/616.  You can see builds in my fork at https://github.com/altendky/miniupnp/pull/1 as long as they aren't being run here.

Draft for:
- [x] Probably need to deal with the ARM builds differently
- [x] Probably want to not use `Chia-Network/actions`
- [x] Consider generating old-style manylinux wheels without the `manylinux_2_12` etc glibc version for pre-pip 20.3 compatibility
  - I would suggest not worrying about this at the moment.  It's not clear that we can properly create the old style presently and projects like `coverage` are using new-style only.  If really needed, maybe we can just do a file copy to the legacy naming style to get them.
- [ ] exclude windows builds
- [ ] look into m1 runners https://github.com/orgs/community/discussions/102846
- [ ] can we run linux arm on the m1 runners without needing qemu?
- [ ] maybe just unpin the cibuildwheel version?

### DIFF
--- a/.github/actions/activate-venv/action.yml
+++ b/.github/actions/activate-venv/action.yml
@@ -1,0 +1,31 @@
+name: "Activate virtual environment"
+description: 'Activate the a virtual environment for future steps.'
+inputs:
+  directories:
+    description: "A list of directories to attempt to activate, terminates on the first present directory.  Uses JSON format."
+    required: true
+    default: '["venv/", ".venv/"]'
+#outputs:
+#  virtual-env:
+#    description: "The VIRTUAL_ENV environment value."
+#    value:
+#  virtual-env-bin:
+#    description: "The path to the virtual environment directory containing the executables."
+#    value:
+runs:
+  using: "composite"
+  steps:
+    - name: Activate virtual environment
+      shell: sh
+      env:
+        INPUT_DIRECTORIES: ${{ inputs.directories }}
+      run: |
+        unset PYTHON
+        for V in 311 3.11 310 3.10 39 3.9 38 3.8 37 3.7 3 ""; do
+          if command -v python$V >/dev/null; then
+            PYTHON=python$V
+            break
+          fi
+        done
+        [ -n "$PYTHON" ] || (echo "Unable to find python" && exit 1)
+        $PYTHON "${GITHUB_ACTION_PATH}/activate_venv.py"
--- a/.github/actions/activate-venv/action.yml
+++ b/.github/actions/activate-venv/action.yml
@@ -1,3 +1,17 @@
+#   Copyright 2024 Chia Network Inc.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
 name: "Activate virtual environment"
 description: 'Activate the a virtual environment for future steps.'
 inputs:
--- a/.github/actions/activate-venv/activate_venv.py
+++ b/.github/actions/activate-venv/activate_venv.py
@@ -1,0 +1,33 @@
+import json
+import os
+import pathlib
+import sys
+
+directories = [
+    pathlib.Path(directory).absolute()
+    for directory in json.loads(os.environ["INPUT_DIRECTORIES"])
+]
+github_env = pathlib.Path(os.environ["GITHUB_ENV"])
+github_path = pathlib.Path(os.environ["GITHUB_PATH"])
+
+for directory in directories:
+    if directory.exists():
+        break
+
+    print(f"No environment found at: {os.fspath(directory)}")
+else:
+    print("No environment found")
+    sys.exit(1)
+
+print(f"Environment found at: {os.fspath(directory)}")
+
+if sys.platform == "win32":
+    bin_scripts_path = directory.joinpath("Scripts")
+else:
+    bin_scripts_path = directory.joinpath("bin")
+
+with github_path.open("a") as file:
+    print(os.fspath(bin_scripts_path), file=file)
+
+with github_env.open("a") as file:
+    print(f"VIRTUAL_ENV={os.fspath(directory)}", file=file)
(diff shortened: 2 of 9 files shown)


miniupnp:
@altendky any update ?

miniupnp:
> Looks like my main concern was the present usage of the chia-network actions. Does it bother you to use these? If so there are a few options. Clone that repo somewhere you control to use, copy just the needed actions into this repo in this PR, just 'hard code' the OS-conditional steps in the two places they are used. Any preference?

well I don't know what chia-network actions are...

please note that the windows python wheels are already built with appveyor : https://pypi.org/project/miniupnpc/

So  I think we will add macOS and linux wheels with your github actions

altendky:
Actions are the main reusable unit in GitHub Actions.  For example, `- uses: actions/checkout@v3` is referencing the official GitHub checkout action.  So the chia-network actions are similar reusable pieces that I wrote for my employer, Chia Network.  While I think they are handy and make things simpler, I understand that other people may prefer to avoid using these company managed tools despite being open source.  They are primarily intended for internal use and while I don't have any intent to significantly change them, breaking changes may happen and we're unlikely to be super careful about breaking external uses.  So, there are a few reasons to avoid using actual direct references to the chia-network actions.

I'll go ahead and suggest that I will make copies of them into this repo.  That makes it easy to update if we find it useful in the future but isolates this project from any changes we may make for internal uses.

I'll adjust this to not build Windows.  That said, I would personally encourage working towards all CI in one place, generally.  I used to leverage Travis, Circle, and AppVeyor to get all the platforms covered.  When Azure Pipelines became available I switched to it since they offered more concurrency and all platforms.  For the past several years I've been working more or less exclusively with GitHub Actions.  Anyways, even if you were to choose to switch I understand it being separate from this PR.  So, as mentioned, I'll exclude the Windows build before considering this PR ready.  It'll be there ready to go though if you change your mind.

altendky:
Do you intend to continue supporting Python 3.7?  I appreciate extended version coverage but just want to make sure before I make accomodations here for it.

altendky:
I have been asked to add our Apache v2 license note and copyright since I'm copying this work out of our repositories.  I'll get it added.  I'm sorry as I understand that this can be an extra thing to think about and we often would be happier to not bother dealing with it.  Let me know if this is an issue and I'll see what I can do to still fulfill the contribution goal here.

miniupnp:
> Do you intend to continue supporting Python 3.7? I appreciate extended version coverage but just want to make sure before I make accomodations here for it.

I don't mind supporting a specific version. I guess the more are supported/built the better.

altendky:
> > Another question, would you like a single artifact you can download with all the generated wheels in it? The new version of the artifact upload action doesn't support adding files from multiple jobs to a single artifact. You have to coalesce them with a download step and then upload a combined artifact. It's easy enough to add this if you want it, otherwise every wheel will be in a separate artifact.
> 
> I don't understand what is the point of a single artifact... What's important is to make the python wheel as widely available as possible. if possible the wheels should be uploaded to [pypi.org/project/miniupnpc](https://pypi.org/project/miniupnpc/)

I'm sorry, I didn't explain clearly enough.

On the 'single artifact' bit, I was not referring to a single wheel.  Rather, if you look at the latest build over in my fork https://github.com/altendky/miniupnp/actions/runs/8191802036?pr=1 you can scroll down on that summary page and see that there is one GitHub Actions artifact per wheel built.  Each of those links will download a `.zip` with one wheel file in it.  I was offering that if you wanted I could make there be an 'everything' artifact and when you download it it would be a single `.zip` containing _all_ of the wheels that had been built.  This just makes it easier to see all the files in one place which is useful at various times.

Setting up automated upload to PyPI for releases can certainly be done as well.  I'll suggest we take that as a separate PR though.

I have a few other items listed in the OP 'draft for' list.  Other than your request that this not build Windows wheels, I think that we could leave those other items for follow up PRs as well.  Just break the work down into more granular pieces.  If you are comfortable with this then I will go ahead and disable the Windows wheel building for now and undraft this for a formal review.

miniupnp:
> > > Another question, would you like a single artifact you can download with all the generated wheels in it? The new version of the artifact upload action doesn't support adding files from multiple jobs to a single artifact. You have to coalesce them with a download step and then upload a combined artifact. It's easy enough to add this if you want it, otherwise every wheel will be in a separate artifact.
> > 
> > 
> > I don't understand what is the point of a single artifact... What's important is to make the python wheel as widely available as possible. if possible the wheels should be uploaded to [pypi.org/project/miniupnpc](https://pypi.org/project/miniupnpc/)
> 
> I'm sorry, I didn't explain clearly enough.
> 
> On the 'single artifact' bit, I was not referring to a single wheel. Rather, if you look at the latest build over in my fork https://github.com/altendky/miniupnp/actions/runs/8191802036?pr=1 you can scroll down on that summary page and see that there is one GitHub Actions artifact per wheel built. Each of those links will download a `.zip` with one wheel file in it. I was offering that if you wanted I could make there be an 'everything' artifact and when you download it it would be a single `.zip` containing _all_ of the wheels that had been built. This just makes it easier to see all the files in one place which is useful at various times.
> 
> Setting up automated upload to PyPI for releases can certainly be done as well. I'll suggest we take that as a separate PR though.

I had understood what you were saying. But I still don't get the point in building a single artifact.
OK, we'll do the automated upload to PyPI later. I guess I would need to do it myself anyway, as I will need to use my PyPI token stored as a github secret.

> 
> I have a few other items listed in the OP 'draft for' list. Other than your request that this not build Windows wheels, I think that we could leave those other items for follow up PRs as well. Just break the work down into more granular pieces. If you are comfortable with this then I will go ahead and disable the Windows wheel building for now and undraft this for a formal review.

altendky:
The single artifact just makes it easier to download them all or even just view all the files.  But that's fine, I'll leave that out.

CI files are the ones that take all the commits what with having to commit and push to test them.  `:]`

I just pushed another commit to exclude Windows, but left the Windows steps in.  I removed some commented out stuff.  The latest run in my fork/pr is available at https://github.com/altendky/miniupnp/actions/runs/8256746251?pr=1 for inspection.  I will undraft this right after I comment since we both are aware of the other steps that would likely be of interest.

Specifically around uploading to PyPI, I should still be able to provide at least a preliminary PR, but yes, you would want to do your own testing etc and do some steps around authorization.  I have worked with the latest system, but I believe there is a new technique for GHA/PyPI interactions that doesn't actually involve your token.  I know we use it at work so I'll look at references and get myself up to speed with how that works.

miniupnp:
It looks good. Do you mind if I "squash and merge" ?

anyway I'm too used to gitlab-ci. Where could I see all the artifacts ?

altendky:
You are welcome to merge however you prefer.

The GitHub Actions artifacts are listed on the summary page such as I linked above, though that is in my repo.  The wheels are inside of those.  I guess either I missed it at first or the workflow has since started running here.  https://github.com/miniupnp/miniupnp/actions/runs/8256746210?pr=640 is the summary link for the latest run in this PR.

PR #643, opened 2023-01-25
Warning when setting index string with sprintf

vonj:
‘%d’ directive output may be truncated writing between 1 and 10 bytes into a region of size 6 [-Werror=format-truncation=]

### DIFF
--- a/miniupnpc/src/upnpc.c
+++ b/miniupnpc/src/upnpc.c
@@ -150,7 +150,7 @@ static void ListRedirections(struct UPNPUrls * urls,
 	printf("PortMappingNumberOfEntries : %u\n", num);*/
 	printf(" i protocol exPort->inAddr:inPort description remoteHost leaseTime\n");
 	do {
-		snprintf(index, 6, "%d", i);
+		snprintf(index, 6, "%d", i % 100000);
 		rHost[0] = '\0'; enabled[0] = '\0';
 		duration[0] = '\0'; desc[0] = '\0';
 		extPort[0] = '\0'; intPort[0] = '\0'; intClient[0] = '\0';


miniupnp:
i < 65536

vonj:
With that modulo we tell the compiler about that too size limiation too. :-)

It's just a fix so one can enable more warnings when building without getting a lot of warnings in the log (or in my case, _stopped_ build because I usually use **-Werror**)

miniupnp:
that's a bug in the compiler static analyzer.

vonj:
I guess you could call it a bug. The data flow analysis did not figure out the variable can never be larger than 65535.
In any case, it's a limitation in the compiler. It may take many years (if ever) before the compiler acquires that ability.

Meanwhile, the format warning on snprintf() is helpful in the general case and it's useful IMHO to have it enabled. Another way to avoid the warning would be to increase the sise of the string slightly.

Are you against addressing these warnings?

miniupnp:
addressing  this warning by making the code less readable is not a good thing.

The best option may be to use a uint16_t for the integer index. But it requires to rewrite the loop.
Or just cast in in the sprintf
```c
snprintf(index, 6, "%hu", (uint16_t)i);
```

vonj:
I'm fine with anything you approve to address the warnings :-)

I can create a pull request for changing to uint16_t, it looks cleanest.

miniupnp:
what compiler gives you the warning ?
I just tried with gcc-12 and clang 13 and -Wextra without any warning

miniupnp:
OK. I have the warning with GCC 10 and GCC 11 but not with GCC 12.

vonj:
I had only tested with gcc 10 I realize now. If I get the idea to try a new pull request to fix warnings, I will make sure to test it against the _latest_ versions of gcc and clang first.  :)

PR #645, opened 2023-01-30
ci: remove .travis.yml

fanquake:
Looks like this is just using GitHub Actions and Appveyor for CI now?


miniupnp:
Yes I believe the free travis-ci is gone forever now :(

Issue #649, opened 2023-03-21
Don't assert default policy in nftables scripts?


HarukaMa:
Currently the nftables init script asserts several default policies:

https://github.com/miniupnp/miniupnp/blob/e439318cf782e30066d430f27a1365e013a5ab94/miniupnpd/netfilter_nft/scripts/nft_init.sh#L21-L23
https://github.com/miniupnp/miniupnp/blob/e439318cf782e30066d430f27a1365e013a5ab94/miniupnpd/netfilter_nft/scripts/nft_init.sh#L47-L49
https://github.com/miniupnp/miniupnp/blob/e439318cf782e30066d430f27a1365e013a5ab94/miniupnpd/netfilter_nft/scripts/nft_init.sh#L57-L59

While those are sensible defaults, setting it as a default action might be disruptive depending on the running system's configuration. (Broke my router, actually. Fortunately it's on a local network.)

Removing the policies should not affect how miniupnpd works. Maybe considering removing them?

HarukaMa:
I'll leave this issue open to let you decide if you still want to change the server's network configuration in a potentially disruptive way by default. You can close this issue if you decide that it's fine.

miniupnp:
as explained by @svenauhagen in #650 you don't have to use the default `nft_init.sh`  and can build your own tables/chains.

`nft_init.sh` is designed to fit the most cases and it is reasonable to drop by default incoming traffic from the internet.

What's the most disruptive, is to install miniupnpd and run its provided init scripts without customizing them and the  whole configuration to your needs first.

PR #650, opened 2023-03-21
Remove default policies for nft chains

HarukaMa:
For #649. Feel free to close this if you think this is not desired.

### DIFF
--- a/miniupnpd/netfilter_nft/scripts/nft_init.sh
+++ b/miniupnpd/netfilter_nft/scripts/nft_init.sh
@@ -20,7 +20,6 @@ cat > /tmp/miniupnpd.nft <<EOF
 table inet $TABLE {
     chain forward {
         type filter hook forward priority 0;
-        policy drop;
 
         # miniupnpd
         jump $CHAIN
@@ -46,7 +45,6 @@ fi
 cat >> /tmp/miniupnpd.nft <<EOF
     chain prerouting {
         type nat hook prerouting priority -100;
-        policy accept;
 
         # miniupnpd
         jump $PREROUTING_CHAIN
@@ -56,7 +54,6 @@ cat >> /tmp/miniupnpd.nft <<EOF
 
     chain postrouting {
         type nat hook postrouting priority 100;
-        policy accept;
 
         # miniupnpd
         jump $POSTROUTING_CHAIN


svenauhagen:
Hi, no you can not remove the default policy. If no policy is set nftables uses accept as policy which means all packets will be accepted even if there is no firewall rule for them. So this will just pass everything.

HarukaMa:
is there a specific reason that miniupnpd *requires* a default reject on forward and accept on input and output? I don't quite think that miniupnpd should mandate what the default policies will be if it's not a hard requirement for it to function.Also the iptables backend is not enforcing this policy as well. Why it's only an issue on nft backend?Sent from my iPhoneOn Mar 22, 2023, at 13:13, sven ***@***.***> wrote:﻿
Hi, no you can not remove the default policy. If no policy is set nftables uses accept as policy which means all packets will be accepted even if there is no firewall rule for them. So this will just pass everything.

svenauhagen:
miniupnpd requires a reject on forward because otherwise the firewall rules generated by miniupnpd would be useless.
All traffic would simply be forwarded no matter what rules are added.
miniupnpd is not creating any input or output chains, you can do that yourself if you like.
postrouting and prerouting are NAT chains and they are set to accept because otherwise all traffic would be dropped that has no DNAT or SNAT rule which is not what you want in a routing setup.
This would only allow traffic that has a matching miniupnpd rule in NAT to go out but you probably also want to allow other traffic to go out into the internet as well not only miniupnpd traffic.
You can always create your own matching ruleset if you like and change the chain names in the miniupnpd config if you have a different setup.

HarukaMa:
> All traffic would simply be forwarded no matter what rules are added.

It doesn't affect the forwarding rules in the prerouting chain, no? Also I've removed the drop policy and miniupnpd is still forwarding incoming traffic without issue, so how is it required?

I just wanted to say that some configurations depend on the default policies to work; it should not be miniupnpd's job to decide what to do with other traffic that is not handled by miniupnpd, except when it's _absolutely required_ for the core functions to work, which I still fail to see why it's required.

Also I have to point it out again that the scripts for iptables backend is _not_ doing this.

svenauhagen:
It still works because the default policy in nftables is accept. This will make the forward rules of miniupnpd useless though since everything will be accepted anyways. Also traffic from the Internet will have unrestricted access to your LAN so you might want to take that into consideration. For DNAT and SNAT there will be no change since miniupnpd sets the policy to accept also. This is to make sure there is a policy and we do not depend on whatever default is set in nftables. 
Iptables and nftables work fundamentally different as that Iptables gives you a pre created set of chains which in nftables you have to create from scratch with tables, chains and policy’s. That is why they are different. You do not have to use the default scripts though you can always write your own rulesets for miniupnpd and nftables. The design is flexible.

HarukaMa:
Yeah, I'm already modifying the scripts locally as the default behavior is really disruptive for me. I think generally you don't want to make potentially disruptive changes to the network policies without notifying the users. (Docker changes the default policy to drop on the forwarding chain as well, and it's frowned upon by many, just FYI.)

I'll close this PR as it seems you are not really interested to change this.

PR #659, opened 2023-06-03
miniupnpc: allow _WIN32_WINNT override

scribam:
It is not the job of a generic build script to decide which version(s) of an OS we are compiling for.

### DIFF
--- a/miniupnpc/CMakeLists.txt
+++ b/miniupnpc/CMakeLists.txt
@@ -40,7 +40,10 @@ if (NOT WIN32)
     target_compile_definitions(miniupnpc-private INTERFACE _NETBSD_SOURCE)
   endif ()
 else ()
-  target_compile_definitions(miniupnpc-private INTERFACE _WIN32_WINNT=0x0501) # XP or higher for getnameinfo and friends
+  set (MINIUPNPC_TARGET_WINDOWS_VERSION "0x0501" CACHE STRING "Minimum target Windows version as hex string") # XP or higher for getnameinfo and friends
+  if (MINIUPNPC_TARGET_WINDOWS_VERSION)
+    target_compile_definitions(miniupnpc-private INTERFACE _WIN32_WINNT=${MINIUPNPC_TARGET_WINDOWS_VERSION})
+  endif ()
 endif ()
 
 if (APPLE)


miniupnp:
> It is not the job of a generic build script to decide which version(s) of an OS we are compiling for.

where should it be defined then ?

scribam:
It should be defined by the parent project.

Example 1: I want to use the default `_WIN32_WINNT` value provided by my Windows SDK
```
project(MyProject)
add_executable(MyProject main.cpp)

add_subdirectory(external/miniupnp/miniupnpc)
target_link_libraries(MyProject PRIVATE libminiupnpc-static)
```

Example 2: I want to specify a common  `_WIN32_WINNT` value across my project (using `add_definitions`)
```
project(MyProject)
add_executable(MyProject main.cpp)

add_definitions(-D_WIN32_WINNT=0x0600)

add_subdirectory(external/miniupnp/miniupnpc)
target_link_libraries(MyProject PRIVATE libminiupnpc-static)
```

Example 3: I want to specify a `_WIN32_WINNT` value for `miniupnpc` (using `target_compile_definitions`)
```
project(MyProject)
add_executable(MyProject main.cpp)

add_subdirectory(external/miniupnp/miniupnpc)
target_compile_definitions(libminiupnpc-static PRIVATE _WIN32_WINNT=0x0600)
target_link_libraries(MyProject PRIVATE libminiupnpc-static)
```

Examples 2 and 3 work with `master` but the defines will have two `_WIN32_WINNT` values : `-D_WIN32_WINNT=0x0501 -D_WIN32_WINNT=0x0600`.

The main issue I have is when I want to build a UWP project using `cmake [...] -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0.22000.0`. In this case, I want to use the default `_WIN32_WINNT` value (which is not `0x0501`). Without this PR, I have to use the following "trick":
```
if(WINDOWS_STORE)
    get_target_property(miniupnpc-private-defs miniupnpc-private INTERFACE_COMPILE_DEFINITIONS)
    list(REMOVE_ITEM miniupnpc-private-defs "_WIN32_WINNT=0x0501")
    set_property(TARGET miniupnpc-private PROPERTY INTERFACE_COMPILE_DEFINITIONS ${miniupnpc-private-defs})
endif()
```

miniupnp:
> It should be defined by the parent project.

so a parent project is mandatory ???

scribam:
Not necessarily. A target (like an executable) in the same project would be ok too.

Without a target, maybe one can set a `CMAKE_C_FLAGS` when using the cmake command line to set the value for `_WIN32_WINNT` but I don't like it.

I have also this option, used by curl:
```
  set(CURL_TARGET_WINDOWS_VERSION "" CACHE STRING "Minimum target Windows version as hex string")
  if(CURL_TARGET_WINDOWS_VERSION)
    add_definitions(-D_WIN32_WINNT=${CURL_TARGET_WINDOWS_VERSION})
    [...]
  endif()
```
https://github.com/curl/curl/blob/9e75932358fa24b8a49b7b13202e866d6e395a79/CMakeLists.txt#L110C4-L112

Then, it can be used like this to force the `_WIN32_WINNT` value instead of using the default one from the SDK:
```
cmake [...] -DCURL_TARGET_WINDOWS_VERSION=0x0501
```

I don't mind keeping `0x0501` as the default value to maintain compatibility.

```
# XP or higher for getnameinfo and friends
set(MINIUPNPC_TARGET_WINDOWS_VERSION "0x0501" CACHE STRING "Minimum target Windows version as hex string")
if(MINIUPNPC_TARGET_WINDOWS_VERSION )
  target_compile_definitions(miniupnpc-private INTERFACE _WIN32_WINNT=${MINIUPNPC_TARGET_WINDOWS_VERSION}) 
endif()
```

What do you think?

miniupnp:
@scribam that's OK if it allows the `_WIN32_WINNT` value to be se from parent project, but defaults to 0x501 when there is no parent project.

scribam:
Done

Issue #663, opened 2023-06-26
Update rule timeout on request (nft ipv6)


sblagodatskikh:
Building with iptables package
```
sblagodatskikh@sblagodatskikh-pc:~$ upnpc -6 -m eth1 -A 2001:777:1:0:ac00:0:13:bf4b 2222 2001:888:0:88:4aee:cff:febb:1c25 6666 TCP 1000
upnpc : miniupnpc library test client. (c) 2005-2014 Thomas Bernard
Go to http://miniupnp.free.fr/ or http://miniupnp.tuxfamily.org/
for more information.
List of UPNP devices found on the network :
 desc: http://[2001:888:0:88::1]:41639/rootDesc.xml
 st: urn:schemas-upnp-org:device:InternetGatewayDevice:1

Found valid IGD : http://[2001:888:0:88::1]:41639/
Local LAN ip address : 2001:888:0:88:4aee:cff:febb:1c25
AddPinhole: ([2001:777:1:0:ac00:0:13:bf4b]:2222 -> [2001:888:0:88:4aee:cff:febb:1c25]:6666) / Pinhole ID = 2
sblagodatskikh@sblagodatskikh-pc:~$
---
~ # ip6tables -nvL MINIUPNPD1
Chain MINIUPNPD1 (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    0     0 ACCEPT     tcp      pppoe_1 *       2001:777:1:0:ac00:0:13:bf4b  2001:888:0:88:4aee:cff:febb:1c25  tcp spt:2222 dpt:6666
~ # 
---
sblagodatskikh@sblagodatskikh-pc:~$ upnpc -6 -m eth1 -A 2001:777:1:0:ac00:0:13:bf4b 2222 2001:888:0:88:4aee:cff:febb:1c25 6666 TCP 10
upnpc : miniupnpc library test client. (c) 2005-2014 Thomas Bernard
Go to http://miniupnp.free.fr/ or http://miniupnp.tuxfamily.org/
for more information.
List of UPNP devices found on the network :
 desc: http://[2001:888:0:88::1]:41639/rootDesc.xml
 st: urn:schemas-upnp-org:device:InternetGatewayDevice:1

Found valid IGD : http://[2001:888:0:88::1]:41639/
Local LAN ip address : 2001:888:0:88:4aee:cff:febb:1c25
[...]

PR #672, opened 2023-11-16
fixes of NFTABLES pinout support

miniupnp:
see #663


svenauhagen:
@miniupnp I will test ipv6 next week for this PR but it looks good. On ipv4 it still works as expected.

miniupnp:
I have rebased on master.
@svenauhagen @juampe thank you for your testing ;)

miniupnp:
it looks like github doesn't see that I have rebased and force pushed the branch

PR #674, opened 2023-12-08
Fix for OpenBSD 7.4

Darkshadow2:
Fixes a change made in pf on OpenBSD 7.4 where one needs to now call `DIOCXEND` for every call to `DIOCGETRULES`. There is now a finite number of tickets, and exceeding this just results in a 'Device busy' error when calling `DIOCGETRULES`. Without this fix, miniupnpd gets to a state where it cannot add/remove any rules.

Here's the relevant part from the 7.4 changelog:
```
Speed up the ioctl(2) request DIOCGETRULE such that pfctl(8)
can retrieve all pf(4) rules from the kernel in linear rather than
in quadratic time. To protect the kernel from memory exhaustion,
userland processes now have to release tickets obtained with
DIOCGETRULES by issuing the new ioctl(2) request DIOCXEND.
In particular, snmpd(8) and systat(1) now do that.
```

### DIFF
--- a/miniupnpd/pf/obsdrdr.c
+++ b/miniupnpd/pf/obsdrdr.c
@@ -1,8 +1,8 @@
-/* $Id: obsdrdr.c,v 1.101 2022/02/19 19:15:24 nanard Exp $ */
+/* $Id: obsdrdr.c,v 1.102 2023/12/07 18:56:32 nanard Exp $ */
 /* vim: tabstop=4 shiftwidth=4 noexpandtab
  * MiniUPnP project
  * http://miniupnp.free.fr/ or https://miniupnp.tuxfamily.org/
- * (c) 2006-2022 Thomas Bernard
+ * (c) 2006-2023 Thomas Bernard
  * This software is subject to the conditions detailed
  * in the LICENCE file provided within the distribution */
 
@@ -73,6 +73,17 @@
 #ifndef USE_PF
 #error "USE_PF macro is undefined, check consistency between config.h and Makefile"
 #else
+
+#if defined(PF_NEWSTYLE) && defined(DIOCXEND)
+#define PF_RELEASETICKETS
+#define release_ticket(device, ticket_num) {\
+	if (ioctl((device), DIOCXEND, &(ticket_num)) < 0) {\
+		syslog(LOG_ERR, "ioctl(dev, DIOCXEND, ...): %m");\
+	}\
+}
+#else
+#define release_ticket(device, ticket_num)	(void)(ticket_num)
+#endif
 
 /* list to keep timestamps for port mappings having a lease duration */
 struct timestamp_entry {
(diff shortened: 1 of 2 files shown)


miniupnp (approved):
I will take the chance to rewrite a bit of the goto code ;)

miniupnp:
@Darkshadow2 could you check that #676 (branch https://github.com/miniupnp/miniupnp/tree/pull-request-674) is still ok for you ?

I added a couple of commits to cleanup the code ;)

Darkshadow2:
@miniupnp yes, it is still ok, though I did need to make some changes. I added a couple of break statements so that the functions would do the same thing as they did before, and in `priv_delete_filter_rule()` you missed a goto statement and the function was always returning -1.

Here's the changes I made to it:
```diff
diff --git a/miniupnpd/pf/obsdrdr.c b/miniupnpd/pf/obsdrdr.c
index 86234ef..58f19d6 100644
--- a/miniupnpd/pf/obsdrdr.c
+++ b/miniupnpd/pf/obsdrdr.c
@@ -984,6 +984,7 @@ get_redirect_rule(const char * ifname, unsigned short eport, int proto,
 			if(timestamp)
 				*timestamp = get_timestamp(eport, proto);
 			r = 0;
+			break;
 		}
 	}
 	release_ticket(dev, tnum);
@@ -1134,6 +1135,7 @@ priv_delete_redirect_rule_check_desc(const char * ifname, unsigned short eport,
 			}
 			remove_timestamp_entry(eport, proto);
 			r = 0;
+			break;
 		}
 	}
 	if (r == -2)
@@ -1174,7 +1176,7 @@ priv_delete_filter_rule(const char * ifname, unsigned short iport,
 	if(ioctl(dev, DIOCGETRULES, &pr) < 0)
 	{
 		syslog(LOG_ERR, "ioctl(dev, DIOCGETRULES, ...): %m");
-		goto error;
+		return -1;
 	}
 	n = pr.nr;
 #ifdef PF_RELEASETICKETS
@@ -1225,7 +1227,7 @@ syslog(LOG_DEBUG, "%2d port=%hu proto=%d addr=%8x",
 	if (r == -2)
 		syslog(LOG_NOTICE, "could not find filter rule to delete iport=%hu addr=%8x", iport, ntohl(iaddr));
 	release_ticket(dev, tnum);
-	return -1;
+	return r;
 #endif
 }
 ```


Apologies for the late reply, I was away from home and then wanted to give the code a few days of testing before answering.

miniupnp:
@Darkshadow2 Thank you for the time you spent !

miniupnp:
@Darkshadow2 can you confirm it is 100% OK now ?

Darkshadow2:
@miniupnp can confirm, it works 100% now.  No problems adding or removing rules, and haven't gotten any errors from the `DIOCGETRULES` calls. Really, haven't had any errors reported at all, so all seems to be working as it was before. :thumbsup:

PR #677, opened 2024-01-01
miniupnpd: Deny secure mode violation with 606 Action not authorized

Self-Hosting-Group:
error and correct 501, 715 and 716 error descriptions.

### DIFF
--- a/miniupnpd/upnpsoap.c
+++ b/miniupnpd/upnpsoap.c
@@ -503,7 +503,7 @@ AddPortMapping(struct upnphttp * h, const char * action, const char * ns)
 	if (strcmp(ext_port, "*") == 0 || eport == 0)
 	{
 		ClearNameValueList(&data);
-		SoapError(h, 716, "Wildcard not permited in ExtPort");
+		SoapError(h, 716, "WildCardNotPermittedInExtPort");
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


miniupnpd/upnpsoap.c
```diff
@@ -479,7 +479,7 @@ AddPortMapping(struct upnphttp * h, const char * action, const char * ns)
 			syslog(LOG_INFO, "Client %s tried to redirect port to %s",
 			       inet_ntoa(h->clientaddr), int_ip);
 			ClearNameValueList(&data);
-			SoapError(h, 718, "ConflictInMappingEntry");
+			SoapError(h, 606, "Action not authorized");
```
miniupnp, line 482: in WANIPConnection-v1, 606 does not exists.

only : 
errorCode errorDescription Description
402 Invalid Args See UPnP Device Architecture section on Control.
501 Action Failed See UPnP Device Architecture section on Control.
715 WildCardNotPermi
ttedInSrcIP
The source IP address cannot be wild-carded
716 WildCardNotPermi
ttedInExtPort
The external port cannot be wild-carded
718 ConflictInMapping
Entry
The port mapping entry specified conflicts with a mapping
assigned previously to another client
724 SamePortValuesRe
quired
Internal and External port values must be the same

PR #678, opened 2024-01-01
miniupnpd: Consider secure mode also for UPnP IGD pinholes

Self-Hosting-Group:
Previously, secure mode was always enabled for pinholes.

### DIFF
--- a/miniupnpd/upnpsoap.c
+++ b/miniupnpd/upnpsoap.c
@@ -1634,7 +1634,7 @@ PinholeVerification(struct upnphttp * h, char * int_ip, unsigned short int_port)
 		strncpy(clientaddr_str, "*ERROR*", sizeof(clientaddr_str));
 	}
 
-	if(memcmp(&h->clientaddr_v6, &result_ip, sizeof(struct in6_addr)) != 0)
+	if(GETFLAG(SECUREMODEMASK) && (memcmp(&h->clientaddr_v6, &result_ip, sizeof(struct in6_addr)) != 0))
 	{
 		syslog(LOG_INFO, "%s: Client %s tried to access pinhole for internal %s and is not authorized",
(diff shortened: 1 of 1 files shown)


miniupnp:
well I have to check in the standard because I think denying creating pinholes for others IP is the recommendation.

miniupnp:
http://upnp.org/specs/gw/UPnP-gw-WANIPv6FirewallControl-v1-Service.pdf 2.6.3.3 p17 : 
> 2.6.3.3 Service Requirements
Before processing the action request, the device MUST apply the chosen security policy, and authenticate
and authorize the control point as required by the security policy. If the control point does not have the
necessary permission to perform the action with the requested parameters, the device MUST return the 606
Action not authorized error code (defined in [DEVICE]).
[IGD2] RECOMMENDS access control requirements and authentication levels to be applied by default for
this action. However, devices MAY choose a different security policy.
In particular, [IGD2] RECOMMENDS that unauthenticated and unauthorized control points are only
allowed to invoke this action with:
• InternalPort value greater than or equal to 1024,
• InternalClient value equals to the control point's IP address.
It is REQUIRED that InternalClient cannot be one of IPv6 addresses used by the gateway.
In cases where the RemoteHost, RemotePort, InternalPort, InternalClient and Protocol are the same than
an existing pinhole, but LeaseTime is different, the device MUST extend the existing pinhole’s lease time
and return the UniqueID of the existing pinhole.

The the current implementation is what is RECOMMENDED by the standard. There is no point in relaxing the implementation.

miniupnp:
the ability to be less secure with ports mappings is to retain compatibility with IGDv1.
As there is no IPv6 pinholes in IGDv1, this is not needed.

miniupnp:
using miniupnpd in a kubernetes cluster ? :thinking: Are you serious ?

PR #679, opened 2024-01-01
miniupnpd: Consider secure mode also for PCP/NAT-PMP and remove the

Self-Hosting-Group:
undocumented pcp_allow_thirdparty configuration option, UPNPPCPALLOWTHIRDPARTY and PCP_ALLOWTHIRDPARTYMASK defines. Secure mode is always enabled for NAT-PMP, as there is no THIRD_PARTY option.


miniupnp:
well I see that option was added in commit c801138c6355f2744f2c09d5e71aeefed1b84527

Why do you want to remove it ?

miniupnp:
the right thing to do is probably to fix the documentation

miniupnp:
> From a user perspective, why have two ways to configure the same thing? It would be more transparent to the user if the secure mode behaves the same as UPnP IGD.

It is not the same thing. PCP is not UPnP IGD.

miniupnp:
> It would probably help this use case too [opnsense/plugins#3727](https://github.com/opnsense/plugins/pull/3727).

It is up to opnsense to the right options to miniupnpd according to what they define.
There is no reason to adapt miniupnpd to how opnsense want to present options to its users.

PR #680, opened 2024-01-01
miniupnpd.conf: Order configuration options and add headings

Self-Hosting-Group:


miniupnpd/miniupnpd.conf
```diff
@@ -33,163 +38,168 @@
...
+#listening_ip=192.168.1.1/24
 
-# Path to the UNIX socket used to communicate with MiniSSDPd
-# If running, MiniSSDPd will manage M-SEARCH answering.
-# default is /var/run/minissdpd.sock
-#minissdpdsocket=/var/run/minissdpd.sock
+# General daemon configuration and enabled protocols
+################################################################################
+# Lease file location
+#lease_file=/var/run/miniupnpd.leases
```
stokito, line 61: The `/var/log/upnp.leases` was changed to the `/var/run/miniupnpd.leases`.
This makes it same as a default in OpenWrt which it good but maybe it's OpenWrt should change its the default instead.
The systemd has an option to create an own service folder under `/var/run/` https://serverfault.com/a/779648

So maybe it would be better to change it to `/var/run/miniupnpd/upnp.leases`.

Anyway the change of defaults must be made in a separate commit and documented in the commit message to make it easier to understand what was changed and create a changelog.

stokito (changes requested):

miniupnpd/miniupnpd.conf
stokito: yes, since it's commented then it should be less risky to change it. Still it should be changed in a separate commit as for me. I feel uncomfortable while trying to understand what was changed and if it's important or not.

miniupnpd/miniupnpd.conf
```diff
@@ -33,159 +38,159 @@
 # Specify STUN UDP port, by default it is standard port 3478.
 #ext_stun_port=3478
 
-# LAN network interfaces IPs / networks
-# There can be multiple listening IPs for SSDP traffic, in that case
-# use multiple 'listening_ip=...' lines, one for each network interface.
-# It can be IP address or network interface name (ie. "eth0")
```
miniupnp, line 39: Why this part has been removed ?

PR #682, opened 2024-01-01
upnpc: Do not show the normally occurring error 713

Self-Hosting-Group:
SpecifiedArrayIndexInvalid when listing port mappings with -l.


miniupnp:
indeed it has already confused a lot of users ;)

miniupnp (approved):

PR #683, opened 2024-01-01
Use the HTTPS-enabled URL of the project website and a small update to

Self-Hosting-Group:
the Python client library description.


miniupnp.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |spec|
-  spec.name = "miniupnp"
-  spec.summary = "Mini UPnP client"
-  spec.homepage = 'http://miniupnp.free.fr/'
+  spec.name = "miniupnpc"
```
miniupnp, line 2: you changed the name, I'm afraid it will cause issues...

PR #686, opened 2024-01-08
miniupnpd: Fix UPnP UDA 1.1 subscribe timeout compliance

Self-Hosting-Group:
SHOULD be greater than or equal to 1800 seconds (30 minutes).
http://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf

### DIFF
--- a/miniupnpd/upnphttp.c
+++ b/miniupnpd/upnphttp.c
@@ -643,8 +643,7 @@ ProcessHTTPSubscribe_upnphttp(struct upnphttp * h, const char * path)
 	       h->req_Timeout);
 	syslog(LOG_DEBUG, "SID '%.*s'", h->req_SIDLen, h->req_buf + h->req_SIDOff);
 #if defined(UPNP_STRICT) && (UPNP_VERSION_MAJOR > 1) || (UPNP_VERSION_MINOR > 0)
-	/*if(h->req_Timeout < 1800) {*/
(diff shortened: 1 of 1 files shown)


miniupnpd/upnphttp.c
```diff
@@ -643,8 +643,7 @@ ProcessHTTPSubscribe_upnphttp(struct upnphttp * h, const char * path)
 	       h->req_Timeout);
 	syslog(LOG_DEBUG, "SID '%.*s'", h->req_SIDLen, h->req_buf + h->req_SIDOff);
 #if defined(UPNP_STRICT) && (UPNP_VERSION_MAJOR > 1) || (UPNP_VERSION_MINOR > 0)
-	/*if(h->req_Timeout < 1800) {*/
-	if(h->req_Timeout == 0) {
```
miniupnp, line 647: Hum I don't remember why I've allowed values between 0 and 1800...

miniupnp:
OK I found where it is required to be 1800+ :
UDA 1.1 4.1.2 p91
> TIMEOUT
> REQUIRED. Field value contains actual duration until subscription expires. Keyword “Second-” followed by an integer (no
space). SHOULD be greater than or equal to 1800 seconds (30 minutes)

miniupnp:
I added a comment.

PR #689, opened 2024-01-08
miniupnpc: Set default lease time to 3600 to fix UPnP IGDv2 compliance

Self-Hosting-Group:
It is RECOMMENDED that a lease time of 3600 seconds would be used as a default value.
http://upnp.org/specs/gw/UPnP-gw-WANIPConnection-v2-Service.pdf

### DIFF
--- a/miniupnpc/src/upnpc.c
+++ b/miniupnpc/src/upnpc.c
@@ -767,7 +767,7 @@ int main(int argc, char ** argv)
 				if (SetRedirectAndTest(&urls, &data,
 						   commandargv[0], commandargv[1],
 						   commandargv[2], commandargv[3],
-						   (commandargc > 4) && is_int(commandargv[4]) ? commandargv[4] : "0",
+						   (commandargc > 4) && is_int(commandargv[4]) ? commandargv[4] : "3600",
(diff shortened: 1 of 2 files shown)


miniupnp:
Well, that's changing the behaviour of the upnpc  command line utility. Not very useful and potentially confusing for users expecting to have "0" as the default value. 

Most people still expect to deal with IGDv1

miniupnp:
> It's all about documentation of the change. The question arises as to whether the old, non-recommended behaviour should be retained or whether it should be changed to the IGDv2 recommendation in the next major release by means of a breaking change in the changelog.

I think you don't understand : most implementations are still IGDv1.
The main reason is that Microsoft Windows is not compatible with IDGv2.

miniupnpd is an exception, there are tricks implemented to fake being IGDv1 when the client is microsoft.
https://github.com/miniupnp/miniupnp/blob/4b0ff229d89d07f40a3e7ac372a592741d954b69/miniupnpd/upnphttp.c#L313

miniupnp:
I see no benefit in this patch and it will break existing usage when 0 is expected.
That is up to the IGD to enforce a maximum lease time.

PR #695, opened 2024-01-10
miniupnpd: Fix default config path, set to /etc/miniupnpd/miniupnpd.conf

Self-Hosting-Group:
to match the path used by the makefiles and distributions. Create a DEFAULT_CONFIG define during configure.

The configuration file installed by the makefiles was not used on Linux before.

### DIFF
--- a/miniupnpd/INSTALL
+++ b/miniupnpd/INSTALL
@@ -126,7 +126,7 @@ https://miniupnp.tuxfamily.org/forum/viewtopic.php?p=4370
 https://github.com/miniupnp/miniupnp/pull/114
 
 =========================== Configuration =============================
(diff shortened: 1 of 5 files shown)


miniupnp:
the Makefile.bsd do install to `/etc/miniupnpd.conf` : 
- https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/Makefile.bsd#L153

indeed the linux Makefiles defines `/etc/miniupnpd/miniupnpd.conf` 
- https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/Makefile.linux#L51
- https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/Makefile.linux_nft

PR #696, opened 2024-01-10
miniupnpc: Set default description to miniupnpc

Self-Hosting-Group:
to match the package name and that of OpenWRT/X-Wrt.
https://github.com/openwrt/packages/blob/master/net/miniupnpc/patches/200-miniupnpc_desc.patch


miniupnp:
I don't really see the point of such change...

miniupnp:
@Self-Hosting-Group Why is that needed or useful ?

stokito:
The libminiupnpc is a library that is used not only by the miniupnpc but also by Transmission, Monero and dozen of other programs.
Changing the user agent may broke integration tests in unexpected places without any benefit.

PR #700, opened 2024-01-15
miniupnpd: Add more options to configure

Self-Hosting-Group:
Add --randomize-urls and --enable-https options and revise help.


miniupnpd/configure
```diff
@@ -29,10 +29,13 @@ case "$argv" in
...
 	--debug) DEBUG=1 ;;
+	--disable-fork) NO_BACKGROUND_NO_PIDFILE=1 ;;
 	--leasefile) LEASEFILE=1 ;;
-	--vendorcfg) VENDORCFG=1 ;;
-	--pcp-peer) PCP_PEER=1 ;;
 	--portinuse) PORTINUSE=1 ;;
+	--pcp-peer) PCP_PEER=1 ;;
+	--randomize-urls) RANDOMIZE_URLS=1 ;;
+	--enable-https) ENABLE_HTTPS=1 ;;
+	--vendorcfg) VENDORCFG=1 ;;
```
miniupnp, line 38: why change order ? it is only confusing and make reviewing harder.

miniupnp:
@Self-Hosting-Group  I don't really get the point of this Pull Request.

miniupnpd/configure
miniupnp: > The options have been intentionally ordered according to how often they are actually used in the distributions. 

I don't understand why it should affect order...

miniupnp:
> > @Self-Hosting-Group I don't really get the point of this Pull Request.
> 
> @miniupnp:
> 
>     * The randomise URLs option has better entropy/security than the random port by default
> 
>     * Allow IGDv2 clients to use a secure transport with HTTPS/securelocation which enables clients to adapt
> 
>     * Update HTTPS_CERTFILE/HTTPS_KEYFILE defaults
> 
>     * ./configure --help more clearly formatted
> 
>     * Add comment: DeviceProtection disabled because incomplete

* The randomise URL is an experimental feature. I does not bring much security. It was add to miniupnpd in order to defeat http://www.filet-o-firewall.com/ but in fact miniupnpd was already not vulnerable thanks to the `Host:` header checking.
* I'm afraid https is not widely supported by IGD control points

miniupnp:
> The attack vector is not a browser, but malware with access to only an HTTP client.
Is that pure speculation or are you describing an existing attack ?

PR #701, opened 2024-01-15
miniupnpc: Exit with 1 if no UPnP devices found with upnp-listdevices

Self-Hosting-Group:


miniupnp:
what is the issue of returning 0 ?

PR #706, opened 2024-01-27
Convert changelogs to markdown

Self-Hosting-Group:
- Rename Changelog.txt to CHANGELOG.md
- Convert changelogs to markdown based on keepachangelog.com
  without types of changes as not used in project and fix capitalisation
- Add links to compare changes for recent releases and unreleased


miniupnp:
Use of markdown is surely an cosmetic improvement, but it will break things like https://miniupnp.tuxfamily.org/files/changelog.php?file=miniupnpd-2.3.4.tar.gz

Issue #710, opened 2024-02-27
UFW integration libmnl mnl_cb_run returned -1


juampe:
I have tried to integrate miniupnp without colliding with ufw and docker. So I have created nftables objects and some configurations related to miniupnpd. I have compiled from this repository and I can test with the latest version. For the miniupnpd configuration, I use the filter and nat tables provided by ufw and create the corresponding objects in after.rules.
I think the problem is that miniupnp is unable to insert the rules for some reason that I can't figure out. I've changed kernels from 6.5.0 to 6.8-rc4 to rule out the problem.
Mainly I could use some information to be able to debug the daemon myself, for example code to print rules over syslog and see what it is trying to do and why it doesn't do it. I have debugged on my own with strace and I can see that it is trying to log the appropriate rules into the nat table. I have tried using, in the code, syslog() and nftnl_rule_snprintf() from libnftnl without much luck.

**Config and logs**

UFW after.rules
```
# BEGIN UFW AND MINIUPNPD
*filter
:miniupnpd - [0:0]
-I FORWARD -j miniupnpd
COMMIT

*nat
:prerouting_miniupnpd - [0:0]
-I PREROUTING -j prerouting_miniupnpd
:postrouting_miniupnpd - [0:0]
-I POSTROUTING -j postrouting_miniupnpd
COMMIT
# END UFW AND MINIUPNPD
```

Nftable object dump related with UFW load
```
table ip filter {
...
       chain FORWARD {
                type filter hook forward priority filter; policy drop;
                counter packets 12440690 bytes 12488369884 jump miniupnpd
                counter packets 13002691 bytes 13100816277 jump DOCKER-USER
                counter packets 13002703 bytes 13100817317 jump DOCKER-ISOLATION-STAGE-1
                oifname "docker0" ct state related,established counter packets 0 bytes 0 accept
                oifname "docker0" counter packets 0 bytes 0 jump DOCKER
                iifname "docker0" oifname != "docker0" counter packets 0 bytes 0 accept
                iifname "docker0" oifname "docker0" counter packets 0 bytes 0 accept
                counter packets 13002813 bytes 13100862318 jump ufw-before-logging-forward
                counter packets 13002813 bytes 13100862318 jump ufw-before-forward
                counter packets 319 bytes 70139 jump ufw-after-forward
                counter packets 319 bytes 70139 jump ufw-after-logging-forward
                counter packets 319 bytes 70139 jump ufw-reject-forward
[...]

PR #711, opened 2024-03-02
NFTables: Add backwards compatibility for seperate IPv4 and IPv6 tables

svenauhagen:
NFtables uses the INET table for NAT and Firewall Rules which combines IPv4 and IPv6. Older systems might not have this option and use the ip/ip6 table instead.

This adds a flag to fall back to the ip/ip6 table style.

Signed-Off-By: Sven Auhagen <sven.auhagen@voleatech.de>


miniupnp:
@ juampe can you have a look at this ? see #710

miniupnp (approved):
it look great !
@juampe does it works for you ?

svenauhagen:
@juampe great, thank you for testing.
@miniupnp can you have a fresh look at the PR?

miniupnp (approved):
looks great !

PR #716, opened 2024-03-13
use macos-14 for linux arm instead of qemu

altendky:


altendky:
I looked around a bit and it appears that Docker support on macOS for GitHub Actions runners is...  not officially there.  Since the QEMU is working for Linux ARM and it doesn't seem to take too terribly long, I think it is ok to just leave it as is for now.

PR #717, opened 2024-03-13
add pypa/gh-action-pypi-publish

altendky:
PyPA has provided a specific action for uploading to PyPI.  https://github.com/pypa/gh-action-pypi-publish  It can be used in various ways but my understanding is that the trusted publisher mode is preferred.  Instructions for the 'authorization' setup are provided at https://docs.pypi.org/trusted-publishers/adding-a-publisher/.

Draft for:
- [x] identifying the preferred method for triggering a publish to PyPI
- [x] implementing the proper condition to replace the presently hard coded `if: false`
- [x] https://github.com/miniupnp/miniupnp/pull/735

### DIFF
--- a/.github/workflows/miniupnpc_wheels.yml
+++ b/.github/workflows/miniupnpc_wheels.yml
@@ -238,6 +238,10 @@ jobs:
           pattern: wheel-*
           path: dist/
 
+      - name: List all artifacts
+        run: |
+          ls -la dist/
+
       - name: Publish distribution to PyPI
         # TODO: identify the desired condition
         if: false
--- a/.github/workflows/miniupnpc_wheels.yml
+++ b/.github/workflows/miniupnpc_wheels.yml
@@ -243,8 +243,7 @@ jobs:
           ls -la dist/
 
       - name: Publish to PyPI
-        # TODO: identify the desired condition
-        if: false
+        if: startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           packages-dir: dist/
--- a/.github/workflows/miniupnpc_wheels.yml
+++ b/.github/workflows/miniupnpc_wheels.yml
@@ -238,7 +238,7 @@ jobs:
           pattern: wheel-*
           path: dist/
 
-      - name: List all artifacts
+      - name: List artifacts
         run: |
           ls -la dist/
 
--- a/.github/workflows/miniupnpc_wheels.yml
+++ b/.github/workflows/miniupnpc_wheels.yml
@@ -226,7 +226,7 @@ jobs:
           if-no-files-found: error
 
   upload:
-    name: ${{ matrix.os.emoji }} 📦 Build ${{ matrix.arch.name }} ${{ matrix.python.major-dot-minor }}
+    name: 🚀 Upload
     runs-on: ubuntu-latest
     needs: build
 
--- a/.github/workflows/miniupnpc_wheels.yml
+++ b/.github/workflows/miniupnpc_wheels.yml
@@ -244,4 +244,6 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           packages-dir: dist/
+          print-hash: true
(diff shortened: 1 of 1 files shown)


altendky:
@miniupnp, do you already have implemented, or have in mind, a mechanism for triggering releases on GitHub that you want to use for the automated upload to PyPI?  If so, I can look into how to create the proper condition to implement that here.

miniupnp:
> @miniupnp, do you already have implemented, or have in mind, a mechanism for triggering releases on GitHub that you want to use for the automated upload to PyPI? If so, I can look into how to create the proper condition to implement that here.

https://github.com/miniupnp/miniupnp/blob/master/appveyor.yml#L102C1-L102C94
`--skip-existing` is what is done ;)

altendky:
I was referring more to what action you want to result in an upload.  Do you want it to publish when you push a tag?  Manually trigger the workflow (dispatch it in Actions speak, I believe)?  Initiate a GitHub release?

Or maybe you are saying you always run the upload step every time the building is done and since you have `--skip-existing` it just doesn't upload anything new until the first run with a bumped version?

miniupnp:
> Or maybe you are saying you always run the upload step every time the building is done and since you have --skip-existing it just doesn't upload anything new until the first run with a bumped version?

That's it ;)

altendky:
Ok, so we have a bit of a special start case here then.  You'll probably want to delete the following two lines whenever you want to do a release.  If we take them out now it will start trying to upload immediately.  It will presumably fail for my external PR.  But...  on `master` it would run if you have the rest setup.

```
        # TODO: identify the desired condition
        if: false
```

I have a couple more things to add.  I set this up over in pytest-twisted over the weekend and found a couple things I had missed here.

altendky:
@miniupnp, does `if: startsWith(github.ref, 'refs/tags')` seem like a reasonable upload condition to you?  I know we talked about how we are using the skip existing option so as to just upload every time, but this seems a bit safer.  For example, without this then the first successful run with access to the secrets (such as you merging this PR) would result in the first release being published.

Also looks like the build needs a fix.  From https://github.com/miniupnp/miniupnp/actions/workflows/miniupnpc_wheels.yml?query=branch%3Amaster it looks like https://github.com/miniupnp/miniupnp/commit/752507ed5191744f248c2a6f3dcc241d6c792de7 broke it, but that is just a comment change so it seems more likely that some not-pinned dependency changed?  Maybe you will recognize the error.

```
        building 'miniupnpc' extension
        creating build/temp.linux-x86_64-cpython-38
        creating build/temp.linux-x86_64-cpython-38/src
        gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Iinclude -I/opt/python/cp38-cp38/include/python3.8 -c src/miniupnpcmodule.c -o build/temp.linux-x86_64-cpython-38/src/miniupnpcmodule.o
        src/miniupnpcmodule.c: In function ‘UPnP_selectigd’:
        src/miniupnpcmodule.c:166:7: error: too few arguments to function ‘UPNP_GetValidIGD’
           r = UPNP_GetValidIGD(self->devlist, &self->urls, &self->data,
               ^~~~~~~~~~~~~~~~
        In file included from src/miniupnpcmodule.c:12:
        include/miniupnpc.h:122:1: note: declared here
         UPNP_GetValidIGD(struct UPNPDev * devlist,
         ^~~~~~~~~~~~~~~~
        error: command '/opt/rh/devtoolset-8/root/usr/bin/gcc' failed with exit code 1
        [end of output]
```

miniupnp:
2b4c0c5ca297c557ef141f3f5fe64f309b319669 has fixed the build.

miniupnp:
> @miniupnp, does if: startsWith(github.ref, 'refs/tags') seem like a reasonable upload condition to you?

thats ok !

miniupnp:
> I think we're good now. Remember to follow the instructions for the 'authorization' setup provided at https://docs.pypi.org/trusted-publishers/adding-a-publisher/.

done.
image

PR #720, opened 2024-03-15
cmake: don't add executable targets when `UPNPC_BUILD_SAMPLE` is `OFF`

jimmy-park:
I think those executable targets should remain in `if(UPNPC_BUILD_SAMPLE)`.


miniupnp:
looks good !

PR #723, opened 2024-04-03
upnpc: Add -f option to delete multiple port redirections

bozheng1988:
Add -f option to delete multiple port redirections.
Use -d to delete batch of port redirections is too slow.


miniupnp:
@bozheng1988 thank you !

PR #733, opened 2024-05-09
appveyor.yml: build for python 3.12 as well and add source distribution file

miniupnp:
image


miniupnp:
python 3.12 lacks modules `setuptools`  and `distutils` : 
image

PR #734, opened 2024-05-09
fix macOS intel wheels

altendky:


altendky:
https://github.com/miniupnp/miniupnp/pull/735 instead

PR #735, opened 2024-05-09
intel: [macos-12]

altendky:
github recently swapped macos-latest to 14 on arm


altendky:
@miniupnp, this is ready for review to fix the broken macos builds.

PR #737, opened 2024-05-09
use macOS 11 as intel deployment target

altendky:


miniupnp:
@altendky could you rebase on master ?

miniupnp:
thank you !

Issue #703, opened 2024-01-17
miniupnpc: Can only open IPv6 ports if XML root device description URL is set


Self-Hosting-Group:
Otherwise, the URL with the IPv6 link-local address is used after discovery, which does not allow ports to be opened for IPv6 GUA addresses.

Also, the exact IPv6 address to initiate the request is not selectable, which is a limitation for client operating systems where multiple temporary IPv6 GUA addresses are typically used.

Control points that have not been authenticated and authorized as defined in IGDv2 SHOULD use their
IPv6 GUA when calling this action.
https://upnp.org/specs/gw/UPnP-gw-WANIPv6FirewallControl-v1-Service.pdf#page=17

Update: Only occurs if the daemon was compiled with UPNP_STRICT defined

miniupnp:
@Self-Hosting-Group If the URL is right (with a IPv6 GUA) then miniupnpc will use the GUA.
Can you provide logs of the problem ?

miniupnp:
The Daemon is advertising it's GUA based URL even when using linklocal Multicast address !
please provide logs of the problem.

miniupnp:
I have to check in the standard what address it is supposed to return in the announcement.

miniupnp:
779a71bb6c939974449cd6f38e2c132f5354d425

miniupnp:
https://openconnectivity.org/upnp-specs/UPnP-arch-DeviceArchitecture-v2.0-20200417.pdf
- p39 : « The URL specified in the LOCATION header field of the M-SEARCH response shall be reachable
by the control point to which the response is directed »
- p112
  - « a) Devices and control points SHALL use only the Link-Local unicast address as the source
address and when specifying a literal IP address in LOCATION URLs in all multicast
messages that are multicast to the Link-Local scope FF02::C for SSDP and FF02::130 for
multicast eventing. Devices and control points SHALL listen on the Link-Local
scope. See RFC 4862 for details of Link-Local addressing. »
  - « f) Devices and control points SHALL use an acquired ULA or GUA in all multicast messages
as the source address and when specifying a literal IP address in LOCATION URLs that
are multicast to the Site-Local scope addresses of either FF05::C or FF05::130 »
  - « j) When a LOCATION URL includes a literal IP address, that IP address SHALL match the
unicast source address used to send the message. When a LOCATION URL includes a
fully qualified domain name (FQDN), the resolution of that FQDN SHALL include the IP
[...]

PR #738, opened 2024-05-10
miniupnpd: Multicast on IPv6 GUA even if UPNP_STRICT is defined

Self-Hosting-Group:
Fix https://github.com/miniupnp/miniupnp/issues/703


miniupnp (changes requested):
```
       /* UDA 1.1 AnnexA and UDA 2.0 only allow/define the use of
       * Link-Local and Site-Local multicast scopes */
```
Just compile without `--strict`  if you want to use Global multicast scope.

miniupnp:
https://openconnectivity.org/upnp-specs/UPnP-arch-DeviceArchitecture-v2.0-20200417.pdf

p112 : 
> g) Devices and control points SHALL NOT send Global scoped, Organization-Local scoped,
or Admin-Local scoped multicast messages.

PR #741, opened 2024-05-10
miniupnpd: add "ext_ip_require_public" option for lower restrictions

rmilecki:
```
Some devices may have private IP address but still receive all external
traffic. In such situations forwarding is still possible and UPnP may be
usable for adding ports mappings.

This case is common with some ISPs that provide they own routers with no
bridge mode support. Users wanting to use their own routers usually just
setup DMZ to redirect all external traffic to their own router.

So far miniupnpd didn't allow adding mappings in such cases. This change
allows modifying that default behaviour using:
ext_ip_require_public=no
```

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1211,6 +1211,7 @@ init(int argc, char * * argv, struct runtime_vars * v)
 
 	/* set initial values */
 	SETFLAG(ENABLEUPNPMASK | SECUREMODEMASK);	/* UPnP and secure mode */
+	SETFLAG(REQUIRE_PUBLIC_EXT_IP);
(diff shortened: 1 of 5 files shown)


rmilecki:
@miniupnp: any opinion on this? Could we have such an option?

miniupnp:
I'm not in favor of adding this possibility. In the UPnP IGD Standard, the router has to know the "public IP Address", so the `ext_ip=` has to be used.
You can also use STUN (options `ext_perform_stun` / `ext_stun_host` / `ext_stun_port`)

miniupnp:
a similar feature was added  in version [2.3.9](https://github.com/miniupnp/miniupnp/releases/tag/miniupnpd_2_3_9).
see #813 / #818

PR #745, opened 2024-05-28
Fix/standardise SOAP/HTTP headers

Self-Hosting-Group:
- Complete 53d4cdd as it only applies to miniupnpd
- Remove `Pragma: no-cache` (standard in HTTP 1.0) and `Cache-Control: no-cache` request headers as they are not used in the UPnP IGD standards and are not implemented in other clients and standardise the `Connection` header

### DIFF
--- a/minissdpd/minissdpd.c
+++ b/minissdpd/minissdpd.c
@@ -472,7 +472,7 @@ SendSSDPMSEARCHResponse(int s, const struct sockaddr * sockname,
 	 *
 	 * have a look at the document "UPnP Device Architecture v1.1 */
 	l = snprintf(buf, sizeof(buf), "HTTP/1.1 200 OK\r\n"
-		"CACHE-CONTROL: max-age=120\r\n"
+		"CACHE-CONTROL: max-age=1800\r\n"
 		/*"DATE: ...\r\n"*/
 		"ST: %.*s\r\n"
 		"USN: %s\r\n"
--- a/miniupnpc-async/miniupnpc-async.c
+++ b/miniupnpc-async/miniupnpc-async.c
@@ -399,7 +399,7 @@ static int upnpc_send_request(upnpc_device_t * p)
 	ssize_t n;
 	static const char reqfmt[] = "GET %s HTTP/1.1\r\n"
 		"Host: %s:%hu\r\n"
-		"Connection: Close\r\n"
+		"Connection: close\r\n"
 		"User-Agent: OS/version UPnP/1.1 MiniUPnPc-async/2.2\r\n"
 		"\r\n";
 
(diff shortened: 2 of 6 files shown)


miniupnp:
I have to check that HTTP/1.0 is never used.

minissdpd/minissdpd.c
```diff
@@ -472,7 +472,7 @@ SendSSDPMSEARCHResponse(int s, const struct sockaddr * sockname,
 	 *
 	 * have a look at the document "UPnP Device Architecture v1.1 */
 	l = snprintf(buf, sizeof(buf), "HTTP/1.1 200 OK\r\n"
-		"CACHE-CONTROL: max-age=120\r\n"
+		"CACHE-CONTROL: max-age=1800\r\n"
```
miniupnp, line 475: why changing from 2 to 30 minutes ?

miniupnpc-async/miniupnpc-async.c
```diff
@@ -399,7 +399,7 @@ static int upnpc_send_request(upnpc_device_t * p)
 	ssize_t n;
 	static const char reqfmt[] = "GET %s HTTP/1.1\r\n"
 		"Host: %s:%hu\r\n"
-		"Connection: Close\r\n"
+		"Connection: close\r\n"
```
miniupnp, line 402: why change from `Close` to `close`  ?

minissdpd/minissdpd.c
miniupnp: https://openconnectivity.org/upnp-specs/UPnP-arch-DeviceArchitecture-v2.0-20200417.pdf
p39 1.3.3 Search response

> CACHE-CONTROL
> Required. Field value shall have the max-age directive (“max-age=”) followed by an integer that specifies the
> number of seconds the advertisement is valid. After this duration, control points should assume the device (or
> service) is no longer available; as long as a control point has received at least one advertisement that is still
> valid from a root device, any of its embedded devices or any of its services, then the control point can assume
> that all are available. The number of seconds should be greater than or equal to 1800 seconds (30 minutes),
> although exceptions are defined in the text above. Specified by UPnP vendor. Other directives shall notbe sent
> and shall be ignored when received

miniupnp:
I have to fix this :
- https://github.com/miniupnp/miniupnp/blob/master/miniupnpc/src/miniupnpc.c#L244
- https://github.com/miniupnp/miniupnp/blob/master/miniupnpc/src/miniwget.c#L487

HTTP version is fixed to 1.1 since ages ;)

miniupnpd/upnpevents.c
```diff
@@ -408,7 +408,6 @@ static void upnp_event_prepare(struct upnp_event_notify * obj)
 		"SID: %s\r\n"
 		"SEQ: %u\r\n"
 		"Connection: close\r\n"
-		"Cache-Control: no-cache\r\n"
```
miniupnp, line 411: why remove ` Cache-control`  ? that is HTTP/1.1

PR #749, opened 2024-05-31
update actions in github workflows

scribam:
- https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
- https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -36,11 +36,11 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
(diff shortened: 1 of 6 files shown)


miniupnp:
@altendky is that OK ?

altendky:
Blech, I left v3 in there?  Sorry.  I don't think this will quite work right as is, but I'll look it over now and provide some details as needed.

miniupnp:
Can I merge the PR now and you fix after if it is needed ? :wink:

altendky:
Oh, I misread at my first glance.  The miniupnpc wheel builds are untouched here.  They were in fact written with v4.  The key difference is that with v3 you could upload multiple times to a single artifact name.  With v4 you can't do that.  The general pattern is to upload to unique names and then download with merging enabled and a wild card that will match all the uniquely named artifacts.  If that isn't relevant any of these places, then it's fine.

altendky:
Yeah, the two uses of `upload-artifact` look fine to me.

PR #753, opened 2024-06-11
miniupnpd: Remove hard port map lifetime code limits for PCP/NAT-PMP

Self-Hosting-Group:
since it already has the same limits by default configuration and it is a "should" not a "must" in the PCP specification. This PR will display a warning/info instead of ignoring the configured limits if outside of the suggestion. This allows the same limits to be configured as UPnP IGD and permit port map tests with a lifetime of e.g. 10s.


miniupnp:
I don't understand why change that

PR #709, opened 2024-02-23
miniupnpd: Add daemon version to XML (child) device descriptions

Self-Hosting-Group:
in addition to the build date and remove dummy strings for model/serial number and some harmonisations.

Issue #698, opened 2024-01-11
miniupnpd: SSDP search response cache-control header, no UPnP UDA 1.0 compliance


Self-Hosting-Group:
The cache-control max-age is set based on notify_interval for SSDP notifications, but is hardcoded for SSDP M-SEARCH responses.

Field value MUST have the max-age directive ("max-age=") followed by an integer that specifies the number of
seconds the advertisement is valid. After this duration, control points SHOULD assume the device (or service) is no longer
available; as long as a control point has received at least one advertisement that is still valid from a root device, any of its
embedded devices or any of its services, then the control point can assume that all are available. The number of seconds
SHOULD be greater than or equal to 1800 seconds (30 minutes), although exceptions are defined in the text above. Specified
by UPnP vendor. Other directives MUST NOT be sent and MUST be ignored when received.
https://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf

Then the default notification interval should probably also be changed to >=900 seconds, so that the effective cache-control header has a max-age of 1800 and is compliant. It might be better to set the max-age to notify_interval * 2 plus 20 seconds and a default notify_interval of >=890.

From the specification:
In addition, the device MUST re-send its advertisements periodically prior to expiration of the duration specified in
the CACHE-CONTROL header field; it is RECOMMENDED that such refreshing of advertisements be done at a randomly-distributed
interval of less than one-half of the advertisement expiration time, so as to provide the opportunity for recovery from lost
advertisements before the advertisement expires, and to distribute over time the advertisement refreshment of multiple devices
on the network in order to avoid spikes in network traffic.

miniupnp:
I have to search the differences between UPnP UDA versions.

[UDA 1.0](http://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.0.pdf) 1.2.3 Discovery: Search: Response (p21) : 
CACHE-CONTROL
Required. Must have max-age directive that specifies number of seconds the advertisement is valid. After this duration,
control points should assume the device (or service) is no longer available. Should be greater than or equal to 1800
seconds (30 minutes), although exceptions are defined in the text above. Specified by UPnP vendor. Integer.

[UDA 1.1](http://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf) 1.3.3 Search response (p34) : 
CACHE-CONTROL
[...]

PR #681, opened 2024-01-01
upnpc: Revise help/usage

Self-Hosting-Group:
Updated help/usage:
```
upnpc: MiniUPnP IGD library test client 2.2.6, (c) 2005-2024 Thomas Bernard, BSD-3-Clause
More information at https://miniupnp.tuxfamily.org or http://miniupnp.free.fr

Usage:
  upnpc [options] -a int_ipv4 int_port ext_port protocol [lease_time] [remote_host]
    Add port mapping (IPv4)
  upnpc [options] -r int_port1 [ext_port1] protocol1 [int_port2 [ext_port2] protocol2] ...
    Add multiple port mappings to the current host (IPv4)
  upnpc [options] -A remote_host remote_port ipv6 port protocol lease_time
    Add pinhole (IPv6, IGDv2 only)
  upnpc [options] -d ext_port protocol [remote_host]
    Delete port mapping (IPv4)
  upnpc [options] -D unique_id
    Delete pinhole (IPv6, IGDv2 only)
  upnpc [options] -l
    List port mappings (IPv4)
  upnpc [options] -L
    List port mappings (IPv4, using GetListOfPortMappings, IGDv2 only)
  upnpc [options] -s
    Get router connection status
  upnpc [options] -S
    Get router firewall status (IGDv2 only)
  upnpc [options] -P
    Get router web interface (presentation) URL
  upnpc [options] -n int_ipv4 int_port ext_port protocol [lease_time] [remote_host]
    Add (any) port mapping to allow IGD to use alternative ext_port (IPv4, IGDv2 only)
  upnpc [options] -N ext_port_start ext_port_end protocol [manage]
    Delete a range of port mappings (IPv4, IGDv2 only)
  upnpc [options] -U unique_id new_lease_time
    Update pinhole (IPv6, IGDv2 only)
  upnpc [options] -K unique_id
    Get number of packets passed through the pinhole (IPv6, IGDv2 only)
  upnpc [options] -C unique_id
    Check if the pinhole is working (IPv6, optional with IGDv2)
  upnpc [options] -G remote_host remote_port ipv6 port protocol
    Get outbound pinhole timeout (IPv6, optional with IGDv2)
Notes:
  Protocol is TCP or UDP. Use "" for any remote_host and 0 for remote_port.
  @ can be used in -a, -A, -n and -G commands to represent IP address of the current host.
Options:
  -e description: Set description of the port mapping (IPv4)
  -6: Use IPv6 instead of IPv4
[...]

PR #754, opened 2024-06-11
Complete proposed changes

Self-Hosting-Group:
The following PRs/issues have not been fully adopted or parts have been overlooked. The following PR tries to fix this.

- Complete: **option enable_natpmp => enable_pcp_pmp** 02da705
  Remove old name still present in documentation/comments
- Complete: #681
  Fix partial help rewrite (remove "redirect", currently more confusing). Add note about optional IGDv2 commands and IPv4/IPv6. Minor wording update/normalise, reduce line width/breaks.
- Complete: #698
  Comment `notify_interval` in configuration, as same value as default and fix help.
- Complete: #709
  Empty dummy values from advertised informal `serialNumber` and `UPC` as IMHO it looks unfinished/unclean not like a long existing project. Remove the change in the root device `modelDescription` to keep the root device white labelled as before.
- Complete: #745
  Remove the `Cache-Control: no-cache` header as it is not used in the IGD specification and by other clients and test tool. Only used in HTTP when using a cache/proxy, which is not possible with IGD. Normally only used in HTTP for GET and not for POST, but in miniupnpc only for POST and not for GET requests. No routers known to need this.

### DIFF
--- a/miniupnpc-async/miniupnpc-async.c
+++ b/miniupnpc-async/miniupnpc-async.c
@@ -694,7 +694,6 @@ static int upnpc_build_soap_request(upnpc_device_t * p, const char * url,
 		"Content-Type: text/xml charset=\"utf-8\"\r\n"
 		"SOAPAction: \"%s#%s\"\r\n"
 		"Connection: close\r\n"
-		"Cache-Control: no-cache\r\n"	/* ??? */
 		"\r\n"
 		"%s";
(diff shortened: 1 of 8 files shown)


miniupnp:
I think now it is almost OK.

Zidane115 (commented):
[](url)

Zidane115 (approved):

PR #756, opened 2024-06-13
miniupnpd: Improve configure - withdrawn

Self-Hosting-Group:
- Use `/etc/os-release` to detect linux dists in configure which also provides `HOME_URL` and is available on most linux distributions, FreeBSD and others
Some distributions have a longer name in `os-release` than with `lsb_release`, so this is just a fallback at the moment, but it allows many lines in configure to be removed later if the longer names are OK.
- Fix configure linux if `--host-os-version` not set to kernelver by setting `HAVE_IP_MREQN=1` by default, this requires linux kernel >= 2.4 from 2001
- miniupnpd: Better default cert file path for experimental HTTPS support


miniupnp (commented):
also we should disable everything when CROSSBUILD=1

PR #766, opened 2024-07-09
enable miniupnpc wheel builds on windows

altendky:
Presently this is just for comparison.  The existing uploaded wheels seem to be missing the actually dll.

image

### DIFF
--- a/.github/workflows/miniupnpc_wheels.yml
+++ b/.github/workflows/miniupnpc_wheels.yml
@@ -136,8 +136,6 @@ jobs:
               matrix: windows
             arch:
               matrix: universal2
-          - os:  # excluding windows entirely as that is presently handled in AppVeyor
-              matrix: windows
           - os:
               matrix: windows
             arch:


altendky:
Looks like they are included when built here.

image

altendky:
Ah, probably because of this hack that got copied over from the Chia builds of miniupnpc.

https://github.com/miniupnp/miniupnp/pull/766/files#diff-1aec4e476d444bcc492fb3510338ac236fb30610cb08b30e959f8eff75e961f7L221
> `7z a dist/*.whl miniupnpc.dll`

altendky:
@miniupnp, do you expect the existing Windows wheels to work on their own without providing a separately built `.dll`?

miniupnp:
> @miniupnp, do you expect the existing Windows wheels to work on their own without providing a separately built `.dll`?

As far as I remember, the .dll is required ;)

miniupnp:
hum It is possible there is a confusion between libminiupnpc.a (static) and miniupnpc.lib + miniupnpc.dll

miniupnp:
https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

As far as I understand, we should get rid of setup.py (and setupmingw32.py) and start from scratch.
But I have yet to find a clear how to ;)

altendky:
Do we (you) want to gate functional Windows wheels on catching up on the past decade of evolution of Python project definitions?  `:]`  Referencing https://github.com/miniupnp/miniupnp/pull/767 really I guess.

But yes, it would certainly be nice to have a `pyproject.toml` perhaps with a local build backend.  I would have to dig back in here.

This PR was mostly just meant to leave behind breadcrumbs on how to enable the Windows build within the GitHub Actions build environment I setup (IIRC).

PR #771, opened 2024-07-31
miniupnpd: Rename config option `force_igd_desc_v1` to `upnp_igd_compat`

### LINKED ISSUE #760: miniupnpd: Improve UPnP IGD compatibility mode configuration option
backward compatible. This will allow new modes and auto-build with `IGD_V2` and `ENABLE_IPV6` defined (more uniform builds), but with IGDv2 disabled by configuration by default, best when PR https://github.com/miniupnp/miniupnp/pull/714 is merged in a future (possibly 2.4.0) release. This would also bring the default miniupnpd builds/configuration into line with how most router projects actually use them.

*Proposed miniupnpd.conf documentation:*

```
# UPnP IGD compatibility mode (default igdv1)
# Available modes: igdv1 (IPv4-only), igdv2 and
# igdv2std (without known incompatible IGDv1 client workaround).
# IGDv1 by default as some IGDv1 clients still incompatible with IGDv2 routers.
# The incompatible Microsoft IGDv1 client always gets IGDv1 except with igd2std.
#upnp_igd_compat=igdv1
```

PR can be provided, where the old configuration option also allowed (`force_igd_desc_v1` with `yes` or `no`).

Self-Hosting-Group:
while still accepting the old name (for backword compatibility) and the new values `igdv1`/`ìgdv2` and still keeping `yes`/`no` as accepted. And improve the documentation of the UPnP IGD compatibility mode.

Fixes #760

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1467,10 +1467,10 @@ init(int argc, char * * argv, struct runtime_vars * v)
 				break;
 #ifdef IGD_V2
 			case UPNPFORCEIGDDESCV1:
-				if (strcmp(ary_options[i].value, "yes") == 0)
+				if ((strcmp(ary_options[i].value, "igdv1") == 0) || (strcmp(ary_options[i].value, "yes") == 0))
 					SETFLAG(FORCEIGDDESCV1MASK);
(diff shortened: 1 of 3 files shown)


miniupnp:
why change to a less descriptive option name ?

PR #775, opened 2024-09-07
Rework buildsystem and add systemd integration

intelfx:
This PR does three things (they are kinda entangled, so it's a single PR; let me know if you want me to split them nonetheless):

- Rework and cleanup the build system with the ultimate goal of extracting the `pkg-config` calls into configure, to avoid quadratic pkg-config invocations (for each library and each source file being compiled);

- Add systemd integration for signaling status and readiness, and update systemd units to make use of the same;

- Install systemd units automatically when built with systemd support.


miniupnpd/.gitignore
```diff
@@ -1,5 +1,4 @@
 *.o
-*.d
```
miniupnp, line 2: We certainly don't want to include dependency files in the git repository.

miniupnpd/Makefile.linux
```diff
@@ -23,13 +23,19 @@
 #
 CONFIG_OPTIONS ?= $(cat .configure.cache)
 CONFIG_OPTIONS += --firewall=iptables
+
+# Make sucks, prove me wrong
+, := ,
+ifeq ($(filter -D_FORTIFY_SOURCE=% -Wp$(,)-D_FORTIFY_SOURCE=%,$(CPPFLAGS) $(CFLAGS)),)
+CPPFLAGS += -D_FORTIFY_SOURCE=2
+endif
+CPPFLAGS += -D_GNU_SOURCE
```
miniupnp, line 32: I won't say it sucks, I'll say it has some quirks :laughing: 

I don't remember in which project I have a variable which contains a single space :sweat_smile:

miniupnp (commented):
thank you for that work, I have to check that everything still works for the other builds (BSD, etc.)

the old build system somehow supports some cross compilation (for OpenWRT for instance), I hope moving some pkg-config stuff from the Makefile to the configure script doesn't break that.

miniupnpd/.gitignore
intelfx: No, this was intended: the dependency files were moved to the `.depend` subdirectory, which was already present in the gitignore file.

intelfx:
> thank you for that work, I have to check that everything still works for the other builds (BSD, etc.)
> 
> the old build system somehow supports some cross compilation (for OpenWRT for instance), I hope moving some pkg-config stuff from the Makefile to the configure script doesn't break that.

It shouldn't (I tried to keep all transformations strictly semantically equivalent), but yup, I'd be grateful if you could recheck the non-Linux and cross builds because I don't have the respective environments set up. (I'll fix the CI failure shortly, it seems I forgot ifdefs in some places.)

miniupnpd/.gitignore
miniupnp: OK, you removed `*.d`  from .gitignore because it is not needed anymore !

miniupnp:
As your commits are atomic, I'm looking at them one by one :)

intelfx:
Hm, I just fixed the ifdefs, but it seems I somehow also broke out-of-tree builds. I'll fix that too.

miniupnp:
I have edited your commit "miniupnpd: configure: generate config.mk consistently" to d68edd90cf64ff8d6de79072204d421889a95b5a

miniupnpd/configure
```diff
@@ -451,23 +524,38 @@ case $OS_NAME in
 			V6SOCKETS_ARE_V6ONLY=`$(find /sbin /bin /usr/sbin /usr/bin -name sysctl) -n net.ipv6.bindv6only`
 			fi
 		fi
+
+		if ldd --version | grep -q GLIBC; then
```
miniupnp, line 528: hum that part wont work for cross builds :(

intelfx: Ideally it needs a compile test, but isn't it exactly the way it was before, just in the Makefile instead of the configure script?

miniupnp: yes, your commit is OK, it was exactly the way it was before.

I'm looking to improve that :)

miniupnp: I guess I should try something like `echo "#include <features.h>" | cc -dM -E -` 
and look for `__GLIBC__` / `__GLIBC_MINOR__`

intelfx: Test-compile something with a call to `clock_gettime()` and see if it links :-)

I'll see if I can stick that into configure.

miniupnp: Or maybe `clock_gettime()` is included in every recent libc :wink: 

it looks like glibc 2.16 is from 2012...

miniupnpd/configure
```diff
@@ -760,6 +760,18 @@ case $OS_FAMILY in
...
 
+# OpenSSL
+case $OS_FAMILY in
+	# OpenWrt disables https server for IGD v2
+	Linux)
+		if pkg_detect libssl; then
+			:
+		else
+			echo "Warning: libssl pkg-config not found!"
+		fi
```
miniupnp, line 787: is there a reason to do that instead of ` if ! pkg_detect libssl ; then`  ?

miniupnpd/Makefile.linux
```diff
@@ -118,23 +118,13 @@ install:	miniupnpd $(SRCDIR)/miniupnpd.8 $(SRCDIR)/miniupnpd.conf \
 	$(INSTALL) $(SRCDIR)/netfilter/ip6tables_removeall.sh $(DESTDIR)$(ETCINSTALLDIR)
 	$(INSTALL) $(SRCDIR)/netfilter/miniupnpd_functions.sh $(DESTDIR)$(ETCINSTALLDIR)
 	$(INSTALL) -m 0644 -b $(SRCDIR)/miniupnpd.conf $(DESTDIR)$(ETCINSTALLDIR)
+	sed -i -e "s/^uuid=[-0-9a-f]*/uuid=$(shell $(UUIDBIN))/" $(DESTDIR)$(ETCINSTALLDIR)/miniupnpd.conf
```
miniupnp: so the generation of the UID is moved to the install target, right ?

miniupnpd/Makefile.linux_nft
```diff
@@ -117,7 +105,7 @@ install:	miniupnpd $(SRCDIR)/miniupnpd.8 $(SRCDIR)/miniupnpd.conf \
 	$(INSTALL) $(SRCDIR)/netfilter_nft/scripts/nft_delete_chain.sh $(DESTDIR)$(ETCINSTALLDIR)
 	$(INSTALL) $(SRCDIR)/netfilter_nft/scripts/miniupnpd_functions.sh $(DESTDIR)$(ETCINSTALLDIR)
 	$(INSTALL) -m 0644 -b $(SRCDIR)/miniupnpd.conf $(DESTDIR)$(ETCINSTALLDIR)
-	sed -i'' -e "s/^uuid=[-0-9a-f]*/uuid=$(UUID)/" $(DESTDIR)$(ETCINSTALLDIR)/miniupnpd.conf
+	sed -i -e "s/^uuid=[-0-9a-f]*/uuid=$(shell $(UUIDBIN))/" $(DESTDIR)$(ETCINSTALLDIR)/miniupnpd.conf
```
miniupnp: 7cd75e8f3ddeb72eb02b0cf239ccb5d976214290 `-i''` is on purpose !

intelfx: Then it needs to be a separate argument, no? Unless I'm fundamentally misunderstanding how BSDs work, the quote removal step runs in the shell and a `''` in the same word as `-i` makes absolutely no difference. To make this actually work as intended by 7cd75e8, there has to be a space between `-i` and `''`.

miniupnpd/configure
intelfx: Purely aesthetical/readability reasons, such that the `if` branch is always the success path (throughout the configure script) and `else` branch is the error path.

miniupnpd/configure
intelfx: The question here is whether `clock_gettime()` is included in `-lc` or whether it needs a `-lrt`.

I can drop this altogether if compatibility with pre-2012 systems is not a concern.

miniupnpd/Makefile.linux_nft
miniupnp: Indeed I thing the space is necessary. I'm going to make a few tests with BSD sed

miniupnpd/configure
miniupnp: According to the [glibc 2.17 announcement](https://sourceware.org/legacy-ml/libc-announce/2012/msg00001.html) I feel that using `clock_gettime()` on glibc 2.16 and before is not worth it : 
> * The `clock_*` suite of functions (declared in <time.h>) is now available
  directly in the main C library.  Previously it was necessary to link with
  -lrt to use these functions.  This change has the effect that a
  single-threaded program that uses a function such as `clock_gettime` (and 
  is not linked with -lrt) will no longer implicitly load the pthreads
  library at runtime and so will not suffer the overheads associated with
  multi-thread support in other code such as the C++ runtime library.

miniupnpd/Makefile.linux_nft
intelfx: Or perhaps just move `-i` to the end of the command line, in front of the file name?

miniupnpd/configure
miniupnp: cleaned there : 8e5e40a8ce9f38400f9779ab74cca7ef7d95ab2a

miniupnpd/Makefile.linux_nft
miniupnp: so as you noticed, `sed -i''` is the same as `sed -i` and works only with Gnu sed.
`sed -i ''` works with BSD sed, but not with Gnu sed. So I'll will continue to assume Linux uses GNU sed and *BSD BSD sed ;)

I guess I should use a variable `SED = sed -i`  / ` SED = sed -i ''` which could be overwritten for strange setups.

intelfx: @miniupnp like I suggested above, could you please test a `sed -e EXPR -i FILE` syntax?

miniupnp: `sed: -I or -i may not be used with stdin`

intelfx: Oh. Yep I've finally taken a look at a BSD sed manpage, the argument after `-i` is the backup file extension, and not just a quirk as I originally thought. So yep it would need two versions of the command line.

miniupnpd/miniupnpd.c
```diff
@@ -1996,6 +2015,16 @@ init(int argc, char * * argv, struct runtime_vars * v)
 		pidfilename = NULL;
 #endif
 
+#ifdef USE_SYSTEMD
+	if (systemd_flag) {
```
miniupnp, line 2019: systemd_flag is never set ...
```c
$ grep -nR systemd_flag *
miniupnpd.c:1185:       int systemd_flag = 0;
miniupnpd.c:2019:       if (systemd_flag) {
```

miniupnpd/miniupnpd.c
```diff
@@ -1192,8 +1203,12 @@ init(int argc, char * * argv, struct runtime_vars * v)
 	{
 		if(0 == strcmp(argv[i], "-h") || 0 == strcmp(argv[i], "--help"))
 			goto print_usage;
-		if(0 == strcmp(argv[i], "-d"))
-			debug_flag = 1;
+		else if(0 == strcmp(argv[i], "-d"))
+			foreground_flag = debug_flag = 1;
+#ifdef USE_SYSTEMD
+		else if(0 == strcmp(argv[i], "-D"))
+			foreground_flag = v->systemd_notify = 1;
```
miniupnp, line 1210: I guess it should be 
```suggestion
			foreground_flag = systemd_flag = 1;
```

intelfx: Yes, indeed, I guess it got out of sync as I was repeatedly refactoring this new code. Thanks!

miniupnp: I'm putting a fixed cherrypick in master anyway

intelfx:
I rebased everything that's left in this PR, added another couple of small fixes to configure and fixed the systemd code.

miniupnpd/miniupnpd.c
intelfx: I changed something else — mind taking a look at it again?

miniupnp:
I think now everything is in master.

Thanks a lot

intelfx:
Yep, looks good. Thanks for the review!

PR #778, opened 2024-09-22
miniupnpc.c: use SOCKET under windows

### LINKED ISSUE #777: Regression in miniupnpc.c (socket type)
The commit https://github.com/miniupnp/miniupnp/commit/5a5b311784c72ff6be113338366d2bdb6b518292 partially reverted changes made 6 years ago, when sockets have got proper type for Windows builds.
Should it be made correct again?
If necessary, a PR could be prepared.

miniupnp:
fixes #777

### DIFF
--- a/miniupnpc/CMakeLists.txt
+++ b/miniupnpc/CMakeLists.txt
@@ -6,7 +6,7 @@ project (miniupnpc
          HOMEPAGE_URL https://miniupnp.tuxfamily.org/
          LANGUAGES C)
 
-set (MINIUPNPC_API_VERSION 18)
+set (MINIUPNPC_API_VERSION 19)
 
 option (UPNPC_BUILD_STATIC "Build static library" TRUE)
 option (UPNPC_BUILD_SHARED "Build shared library" TRUE)
--- a/miniupnpc/Changelog.txt
+++ b/miniupnpc/Changelog.txt
@@ -1,5 +1,9 @@
-$Id: Changelog.txt,v 1.269 2024/07/27 11:53:44 nanard Exp $
+$Id: Changelog.txt,v 1.270 2025/01/10 22:57:19 nanard Exp $
 miniUPnP client Changelog.
+
+2025/01/10:
+  Change simpleUPnPcommand() prototype to remove unused first argument
+  Increments API_VERSION to 19
 
 2024/07/27:
   Add #define for UPNP_GetValidIGD() return values
--- a/miniupnpc/Makefile
+++ b/miniupnpc/Makefile
@@ -84,7 +84,7 @@ ifneq (, $(findstring sun, $(OS))$(findstring solaris, $(OS)))
 endif
 
 # APIVERSION is used to build SONAME
-APIVERSION = 18
+APIVERSION = 19
 
 SRCS = $(wildcard $(SRCDIR)/*.c)
 
--- a/miniupnpc/apiversions.txt
+++ b/miniupnpc/apiversions.txt
@@ -1,6 +1,11 @@
-$Id: apiversions.txt,v 1.9 2016/01/24 17:24:36 nanard Exp $
+$Id: apiversions.txt,v 1.12 2025/01/10 22:57:21 nanard Exp $
 
 Differences in API between miniUPnPc versions
+
+API version 19
+  change simpleUPnPcommand() prototype
+  updated macro :
+    #define MINIUPNPC_API_VERSION 19
 
 API version 18
   change UPNP_GetValidIGD() prototype and return values
--- a/miniupnpc/include/miniupnpc.h
+++ b/miniupnpc/include/miniupnpc.h
@@ -3,7 +3,7 @@
  * Project: miniupnp
  * http://miniupnp.free.fr/ or https://miniupnp.tuxfamily.org/
  * Author: Thomas Bernard
- * Copyright (c) 2005-2024 Thomas Bernard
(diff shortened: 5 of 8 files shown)


miniupnp:
@irwir does it fixes your problem ?

irwir:
This complicates code but does not resolve the isssue: 32-bit signed integer vs unsigned 32 and 64-bit requires type casts with potential loss of data.
Proper fix should restore the declaration to
`simpleUPnPcommand(SOCKET s, const char * url, const char * service,`
and calls should use `INVALID_SOCKET` constant instead of `-1`.

miniupnp:
@irwir the prototype has been unchanged for at least 13 years and never has been a problem.
https://github.com/miniupnp/miniupnp/blame/master/miniupnpc/include/miniupnpc.h#L40

```c
char *
simpleUPnPcommand(int, const char *, const char *,
                  const char *, struct UPNParg *,
                  int *);
```

winsock2.h defines SOCKET as follow :
`typedef u_int   SOCKET;`  so using `int` and casting works.

irwir:
> the prototype has been unchanged for at least 13 years and never has been a problem.

For the last 6 years it was SOCKET; the link to the commit was given in my issue.

> winsock2.h defines SOCKET as follow :
> `typedef u_int SOCKET;` so using `int` and casting works.

Where you have got this header?
The correct definition in VS2022 is 
`typedef UINT_PTR        SOCKET;`
This is 64-bit value in 64-bit builds.

miniupnp:
> > the prototype has been unchanged for at least 13 years and never has been a problem.
> 
> For the last 6 years it was SOCKET; the link to the commit was given in my issue.

I think there is a confusion between `simpleUPnPcommand` and `simpleUPnPcommand2` which was using a `SOCKET` argument.
When I removed  `simpleUPnPcommand2` in c4bd33b409e102ad2183dd89e7aa037a24dd5e30, I inadvertently created an inconsistency between the  `simpleUPnPcommand` prototype in the .c and in the .h :disappointed: 

> > winsock2.h defines SOCKET as follow :
> > `typedef u_int SOCKET;` so using `int` and casting works.
> 
> Where you have got this header? The correct definition in VS2022 is `typedef UINT_PTR SOCKET;` This is a 64-bit value in 64-bit builds.

From mingw32.
That's possible that windows is using 64bits in 64 bits builds. But we are not using any negative value in `simpleUPnPcommand`. No one should call `simpleUPnPcommand` with -1 or INVALID_SOCKET...

To fix this issue for good the miniupnpc API should be modified to use `SOCKET`  for win32 only.
in POSIX, the type for socket file descriptor is int.
I really don't understand why microsoft didn't follow the standard in WinSock2 :sob:

irwir:
> inconsistency between the `simpleUPnPcommand` prototype in the .c and in the .h

The first argument is used as a socket, hence it must be `SOCKET`.
Troubles began when directory structure has been changed and `miniupnpc_socketdef.h` was not included whenever required. Otherwise, there was my old PR, which was manually reimplement by you and everything was good for quite a while. The required additional changes are small.

> From mingw32.

The correct file for mingw is [_socket_types.h](https://github.com/mingw-w64/mingw-w64/blob/28e642fcdc9e7ce64bc3d4456c6ba143bb449e70/mingw-w64-headers/include/psdk_inc/_socket_types.h)

miniupnp:
In fact the 1st argument should be removed. It is not used internally in miniupnpc.
But I don't know how many external projects are using `upnpDiscoverDevice()` directly :(

irwir:
> In fact the 1st argument should be removed. It is not used internally in miniupnpc.

Things may change in next versions, but this is the current code:
https://github.com/miniupnp/miniupnp/blob/a71563298745bb94bdc71b1493db3c08a532acb6/miniupnpc/src/miniupnpc.c#L200-L203
`connecthostport` returns SOCKET type, and `int s` is no big enough for 64-bit value.

irwir:
The suggestion would be to add local `SOCKET s;` variable.
Then the first parameter could be removed.
Othewise, the parameter could be renamed, for example, to `dummy` and `(unsigned)dummy;` statement has to be added to prevent compiler warnings.

miniupnp:
> > In fact the 1st argument should be removed. It is not used internally in miniupnpc.
> 
> Things may change in next versions, but this is the current code:
> 
> https://github.com/miniupnp/miniupnp/blob/a71563298745bb94bdc71b1493db3c08a532acb6/miniupnpc/src/miniupnpc.c#L200-L203
> 
> `connecthostport` returns `SOCKET` type, and `int s` is not big enough for a 64-bit value.

that is exactly what this PR is doing : declaring s as `SOCKET`.

irwir:
> that is exactly what this PR is doing : declaring s as `SOCKET`.

It does, but in a strange ways.
First, explicit use of the 1st parameter that might get dropped in future releases.
Second, the major point of intorducing the SOCKET and related macros was uniform code in WIndows and *nix systems. But this PR again requires a conditional compilation block. If you always use -1 or invalid socket, there is no reason to assign and to check for invalid socket at the line 200.
Finally, a side effect. If a valid socket was passed as the 1st parameter, it might get closed in certain cases. This is not right for 64-bit SOCKET which was passed as an integer.

miniupnp:
Yes I see that passing a socket in first argument is indeed bugged. 
I guess I had written this function in order to be able to reuse the same socket for several successive SOAP methods calls, but I guess it was never used and the code will probably not work well...

PR #780, opened 2024-09-23
miniupnpd: Small Makefile.bsd, man page fixes and IGD config improvements

Self-Hosting-Group:
- Fix Makefile.bsd man dir creation
  Follow Linux makefiles and allow removal of downstream patch in FreeBSD ports
- Fix manual page, harmonise and remove obsolete note
- Allow and ignore UPnP IGD compatibility mode runtime option if compiled without IGDv2 support
  Current behaviour: daemon exits with error 1. This allows pre-configuration of IGDv1 mode (for client compatibility) in a UI frontend (configuration) package, regardless of whether the daemon package is already compiled with IGDv2 support, for smoother/independent updates of separate packages as in e.g. OPNsense/pfSense

### DIFF
--- a/miniupnpd/INSTALL
+++ b/miniupnpd/INSTALL
@@ -173,9 +173,9 @@ allowed, so it is a good practice to have a "catch all" deny permission
 rule at the end of your permission ruleset.
 Sample permission ruleset :
 allow 4662-4672 192.168.1.34/32 4662-4672
-deny 0-65535 192.168.1.34/32 0-65535
+deny 1-65535 192.168.1.34/32 1-65535
 allow 1024-65535 192.168.1.0/24 1024-65535
-deny 0-65535 0.0.0.0/0 0-65535
+deny 1-65535 0.0.0.0/0 1-65535
 With this ruleset, redirections are allowed only for host on the subnet
 192.168.1.0/255.255.255.0 for the ports 1024 or above. There is an exception
 for the host 192.168.1.34 for which only redirections from/to port 4662 to
--- a/miniupnpd/Makefile.bsd
+++ b/miniupnpd/Makefile.bsd
@@ -130,7 +130,7 @@ MANPREFIX ?= $(PREFIX)
 .if $(MANPREFIX) == ""
 MANPREFIX = /usr/share
 .endif
-INSTALLMANDIR = $(MANPREFIX)/man
+INSTALLMANDIR = $(MANPREFIX)/man/man8
 
 all:	$(EXECUTABLES)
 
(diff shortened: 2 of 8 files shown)


miniupnpd/Makefile.bsd
```diff
@@ -153,7 +153,7 @@ install:	miniupnpd genuuid
 	$(INSTALL) -d $(DESTDIR)$(INSTALLETCDIR)
 	$(INSTALL) -b miniupnpd.conf $(DESTDIR)$(INSTALLETCDIR)
 	$(INSTALL) -d $(DESTDIR)$(INSTALLMANDIR)
-	$(INSTALL) -m 644 $(SRCDIR)/miniupnpd.8 $(DESTDIR)$(INSTALLMANDIR)/man8/miniupnpd.8
+	$(INSTALL) -m 644 $(SRCDIR)/miniupnpd.8 $(DESTDIR)$(INSTALLMANDIR)/miniupnpd.8
```
miniupnp, line 156: what is the point of this change ?

miniupnpd/INSTALL
```diff
@@ -173,9 +173,9 @@ allowed, so it is a good practice to have a "catch all" deny permission
 rule at the end of your permission ruleset.
 Sample permission ruleset :
 allow 4662-4672 192.168.1.34/32 4662-4672
-deny 0-65535 192.168.1.34/32 0-65535
+deny 1-65535 192.168.1.34/32 1-65535
 allow 1024-65535 192.168.1.0/24 1024-65535
-deny 0-65535 0.0.0.0/0 0-65535
+deny 1-65535 0.0.0.0/0 1-65535
```
miniupnp, line 178: whats is the point of this change ?

miniupnpd/Makefile.bsd
miniupnp: to fix the directory creation, the line
```
$(INSTALL) -d $(DESTDIR)$(INSTALLMANDIR)
```
should be changed to
```
$(INSTALL) -d $(DESTDIR)$(INSTALLMANDIR)/man8
```
see https://github.com/freebsd/freebsd-ports/blob/main/net/miniupnpd/files/patch-Makefile.bsd

miniupnpd/INSTALL
miniupnp: :facepalm: the BSD sockets use port 0 for "any" port.
your modification is adding confusion.

The purpose of the "catch all" deny rule is to deny everything. Allowing request for port 0 is a big risk of mix-up.

PR #796, opened 2024-12-30
OpenBSD fix for removal of expired rules & miniupnpd ssdp fixes

Darkshadow2:
- miniupnpd/pf/obsdrdr.c
  Fixes rules not being removed from the pf anchor when they expire
    Apparently, it changed at some point and calling DIOCGETRULE
    always starts from the beginning of the list on OpenBSD 7.5
    (and possibly earlier - not entirely sure when this changed)
    This was the only place I found that needed this change, all
    other calls to DIOCGETRULE are already being run in a loop
    and work as intended

- miniupnpd ssdp changes
  Fixes a couple of issues with miniupnpd being the SSDP listener/server
  - miniupnpd/minissdp.c
      Fixes a runtime warning
  - miniupnpd/bsd/getroute.c
      Fixes routes not being found. Note that this changes it
      to be the same as the equivalent code in minissdp itself

### DIFF
--- a/miniupnpd/bsd/getroute.c
+++ b/miniupnpd/bsd/getroute.c
@@ -65,7 +65,7 @@ get_src_for_route_to(const struct sockaddr * dst,
 	rtm.rtm_flags = RTF_UP;
 	rtm.rtm_version = RTM_VERSION;
 	rtm.rtm_seq = 1;
-	rtm.rtm_addrs = RTA_DST | RTA_IFA | RTA_IFP;	/* pass destination address, request source address & interface */
+	rtm.rtm_addrs = RTA_DST;
 	memcpy(m_rtmsg.m_space, dst, l);
 #if !defined(__sun)
 	((struct sockaddr *)m_rtmsg.m_space)->sa_len = l;
--- a/miniupnpd/bsd/getroute.c
+++ b/miniupnpd/bsd/getroute.c
@@ -16,6 +16,9 @@
 #ifdef AF_LINK
 #include <net/if_dl.h>
 #endif
+#ifdef __OpenBSD__
+#include <sys/param.h>
+#endif
 
 #include "config.h"
 #include "../upnputils.h"
@@ -65,7 +68,11 @@ get_src_for_route_to(const struct sockaddr * dst,
 	rtm.rtm_flags = RTF_UP;
 	rtm.rtm_version = RTM_VERSION;
 	rtm.rtm_seq = 1;
+#if defined(OpenBSD) && OpenBSD >= 201911
 	rtm.rtm_addrs = RTA_DST;
+#else
+	rtm.rtm_addrs = RTA_DST | RTA_IFA | RTA_IFP;	/* pass destination address, request source address & interface */
+#endif
 	memcpy(m_rtmsg.m_space, dst, l);
 #if !defined(__sun)
 	((struct sockaddr *)m_rtmsg.m_space)->sa_len = l;
--- a/miniupnpd/pf/obsdrdr.c
+++ b/miniupnpd/pf/obsdrdr.c
@@ -1471,11 +1471,13 @@ get_redirect_rule_by_index(int index,
 	}
 #else /* USE_LIBPFCTL */
 	pr.nr = index;
-	if(ioctl(dev, DIOCGETRULE, &pr) < 0)
-	{
-		syslog(LOG_ERR, "ioctl(dev, DIOCGETRULE): %m");
-		goto error;
-	}
+	do {
+		if(ioctl(dev, DIOCGETRULE, &pr) < 0)
+		{
+			syslog(LOG_ERR, "ioctl(dev, DIOCGETRULE): %m");
+			goto error;
+		}
+	} while ( pr.nr != (unsigned int)index );
 #endif /* USE_LIBPFCTL */
 	*proto = RULE.proto;
 #ifdef __APPLE__
--- a/miniupnpd/minissdp.c
+++ b/miniupnpd/minissdp.c
@@ -220,7 +220,11 @@ OpenAndConfSSDPReceiveSocket(int ipv6)
 #endif /* IP_PKTINFO */
 #if defined(ENABLE_IPV6) && defined(IPV6_RECVPKTINFO)
 	if(ipv6) {
+#ifdef IPPROTO_IPV6
+		if(setsockopt(s, IPPROTO_IPV6, IPV6_RECVPKTINFO, &on, sizeof(on)) < 0)
+#else
 		if(setsockopt(s, IPPROTO_IP, IPV6_RECVPKTINFO, &on, sizeof(on)) < 0)
+#endif
 		{
 			syslog(LOG_WARNING, "setsockopt(udp, IPV6_RECVPKTINFO): %m");
 		}


miniupnpd/pf/obsdrdr.c
```diff
@@ -1471,11 +1471,13 @@ get_redirect_rule_by_index(int index,
...
-		syslog(LOG_ERR, "ioctl(dev, DIOCGETRULE): %m");
-		goto error;
-	}
+	do {
+		if(ioctl(dev, DIOCGETRULE, &pr) < 0)
+		{
+			syslog(LOG_ERR, "ioctl(dev, DIOCGETRULE): %m");
+			goto error;
+		}
+	} while ( pr.nr != (unsigned int)index );
```
miniupnp, line 1480: I'm confused : is there a risk of infinite loop ?
and why pr.nr would be different in the first place ?

miniupnp: https://man.openbsd.org/pf has not changed : 

> [DIOCGETRULE](https://man.openbsd.org/pf#DIOCGETRULE) _struct pfioc_rule *pr_
>     Get a _rule_ by its number _nr_ using the ticket obtained through a preceding `DIOCGETRULES` call. If action is set to `PF_GET_CLR_CNTR`, the per-rule statistics on the requested _rule_ are cleared.

miniupnpd/bsd/getroute.c
```diff
@@ -65,7 +65,7 @@ get_src_for_route_to(const struct sockaddr * dst,
 	rtm.rtm_flags = RTF_UP;
 	rtm.rtm_version = RTM_VERSION;
 	rtm.rtm_seq = 1;
-	rtm.rtm_addrs = RTA_DST | RTA_IFA | RTA_IFP;	/* pass destination address, request source address & interface */
+	rtm.rtm_addrs = RTA_DST;
```
miniupnp, line 68: with this modification it fails under FreeBSD 13.4

```
testgetroute 3601 - - calling get_src_for_route_to(0x82103c3f0, NULL, NULL, 0x82103c404)
testgetroute 3601 - - read l=200 seq=1 pid=3601   sizeof(struct rt_msghdr)=152
testgetroute 3601 - - offset=  0 type= 1 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3601 - - offset= 16 type= 2 sa_len=16 sa_family=2 192.168.11.254:0
testgetroute 3601 - - offset= 32 type= 4 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3601 - - get_src_for_route_to() returned -1
testgetroute 3601 - - calling get_src_for_route_to(0x82103c3f0, 0x82103c3f4, 0x82103c3e8(4), 0x82103c404)
testgetroute 3601 - - read l=200 seq=1 pid=3601   sizeof(struct rt_msghdr)=152
testgetroute 3601 - - offset=  0 type= 1 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3601 - - offset= 16 type= 2 sa_len=16 sa_family=2 192.168.11.254:0
testgetroute 3601 - - offset= 32 type= 4 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3601 - - get_src_for_route_to() returned -1
```

without the modification : 
```
testgetroute 3623 - - calling get_src_for_route_to(0x8207d65e0, NULL, NULL, 0x8207d65f4)
testgetroute 3623 - - read l=272 seq=1 pid=3623   sizeof(struct rt_msghdr)=152
testgetroute 3623 - - offset=  0 type= 1 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3623 - - offset= 16 type= 2 sa_len=16 sa_family=2 192.168.11.254:0
testgetroute 3623 - - offset= 32 type= 4 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3623 - - offset= 48 type=16 sa_len=56 sa_family=18 index=1 type=6 em0:x.x.x.x.x.x
testgetroute 3623 - - offset=104 type=32 sa_len=16 sa_family=2 192.168.11.144:0
testgetroute 3623 - - get_src_for_route_to() returned 0
testgetroute 3623 - - index=1
testgetroute 3623 - - calling get_src_for_route_to(0x8207d65e0, 0x8207d65e4, 0x8207d65d8(4), 0x8207d65f4)
testgetroute 3623 - - read l=272 seq=1 pid=3623   sizeof(struct rt_msghdr)=152
testgetroute 3623 - - offset=  0 type= 1 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3623 - - offset= 16 type= 2 sa_len=16 sa_family=2 192.168.11.254:0
testgetroute 3623 - - offset= 32 type= 4 sa_len=16 sa_family=2 0.0.0.0:0
testgetroute 3623 - - offset= 48 type=16 sa_len=56 sa_family=18 index=1 type=6 em0:x.x.x.x.x.x
testgetroute 3623 - - offset=104 type=32 sa_len=16 sa_family=2 192.168.11.144:0
testgetroute 3623 - - get_src_for_route_to() returned 0
testgetroute 3623 - - src=192.168.11.144:0
testgetroute 3623 - - index=1
```

miniupnp: same behavior with OpenBSD 4.5... ;)

miniupnp: see e5d30a1f4bff488292ff0575286adce536ec5dce eb5f179c9f62989274d1c761156192d5b37b4a4a

miniupnp: https://github.com/miniupnp/miniupnp/pull/164

miniupnpd/pf/obsdrdr.c
Darkshadow2: No, the issue is that it will always return rules in order. So, if the loop is not there it will always return the first rule when it is called (setting pr.nr before has no effect). The loop is there so that it will actually return the requested index.

Darkshadow2: As an example, say there are 4 rules in the anchor. The very first rule is one that has days left before it expires (since a week is the max). The other rules should expire within the next hour. Calling this will _always_ return the first rule (pr.nr == 0) for all calls to `get_redirect_rule_by_index()` without this change (on OpenBSD 7.5, at least). This breaks the other rules being cleaned when expired, as only the very first rule in the list will ever be checked.

miniupnpd/bsd/getroute.c
Darkshadow2: Hmm, interesting. It has actually been failing to work under OpenBSD since 6.something (I don't really remember which version). I just decided to add it here because the equivalent code in minissdpd was the same and I figured that worked everywhere.

This is what I get without the modification:
```
testgetroute: calling get_src_for_route_to(0x6faafa06b6e0, NULL, NULL, 0x6faafa06b6fc)
testgetroute: write: Invalid argument
testgetroute: get_src_for_route_to() returned -1
testgetroute: calling get_src_for_route_to(0x6faafa06b6e0, 0x6faafa06b6e4, 0x6faafa06b6d8(4), 0x6faafa06b6fc)
testgetroute: write: Invalid argument
testgetroute: get_src_for_route_to() returned -1
```

And with the modification:
```
testgetroute: calling get_src_for_route_to(0x7816cb21b150, NULL, NULL, 0x7816cb21b16c)
testgetroute: read l=208 seq=1 pid=85970   sizeof(struct rt_msghdr)=96
testgetroute: offset=  0 type= 1 sa_len=16 sa_family=2 10.0.3.5:0
testgetroute: offset= 16 type= 2 sa_len=32 sa_family=18 index=1 type=6 y.y.y.y.y.y
testgetroute: offset= 48 type= 4 sa_len=16 sa_family=2 255.255.255.255:0
testgetroute: offset= 64 type=16 sa_len=32 sa_family=18 index=1 type=6 ix0:x.x.x.x.x.x
testgetroute: offset= 96 type=32 sa_len=16 sa_family=2 10.0.0.1:0
testgetroute: get_src_for_route_to() returned 0
testgetroute: index=1
testgetroute: calling get_src_for_route_to(0x7816cb21b150, 0x7816cb21b154, 0x7816cb21b148(4), 0x7816cb21b16c)
testgetroute: read l=208 seq=1 pid=85970   sizeof(struct rt_msghdr)=96
testgetroute: offset=  0 type= 1 sa_len=16 sa_family=2 10.0.3.5:0
testgetroute: offset= 16 type= 2 sa_len=32 sa_family=18 index=1 type=6 y.y.y.y.y.y
testgetroute: offset= 48 type= 4 sa_len=16 sa_family=2 255.255.255.255:0
testgetroute: offset= 64 type=16 sa_len=32 sa_family=18 index=1 type=6 ix0:x.x.x.x.x.x
testgetroute: offset= 96 type=32 sa_len=16 sa_family=2 10.0.0.1:0
testgetroute: get_src_for_route_to() returned 0
testgetroute: src=10.0.0.1:0
testgetroute: index=1
```

miniupnpd/pf/obsdrdr.c
Darkshadow2: There should be no functional change for any system where setting pr.nr before calling DIOCGETRULE works. In that case the test should fail and it will never loop.

miniupnpd/bsd/getroute.c
Darkshadow2: This is not going to be easy to implement, then. A way to detect which is needed needs to be devised, and it' not simply going to be 'If this is OpenBSD or not.' I will try to see if I can think of a way, but if not I'll remove this portion from this pull request.

Darkshadow2: Found this changed in OpenBSD 6.6, added a commit that detects if it's running on OpenBSD 6.6 later and does this modification. Should be all good now

miniupnpd/pf/obsdrdr.c
Darkshadow2: Oh I missed your question about if it would be an infinite loop. It shouldn't, if `DIOCGETRULE` returns an error (like getting to the end of the list), the `goto` statement will break it out of the loop.

miniupnp: What's strange is that the current OpenBSD man states that setting pr.nr before calling DIOCGETRULE works.
Well at least it was the way it worked when they wrote the man and they haven't changed the man.

That looks like a bug in either OpenBSD man or the DIOCGETRULE implementation ;)

Darkshadow2: I'll agree with you there. A bug in one or the other. This is also why I'm not sure when it changed. Possibly 7.4, when other pf stuff was changed (like the addition of DIOCXEND).

Darkshadow2: 7.4 may indeed be when it changed. I was looked at #674 where support for `DIOCXEND` was added, and I had this from the 7.4 changelog:

```
Speed up the ioctl(2) request DIOCGETRULE such that pfctl(8)
can retrieve all pf(4) rules from the kernel in linear rather than
in quadratic time. To protect the kernel from memory exhaustion,
userland processes now have to release tickets obtained with
DIOCGETRULES by issuing the new ioctl(2) request DIOCXEND.
In particular, snmpd(8) and systat(1) now do that.
```

That bit about speeding up `DIOCGETRULE` wasn't really relevant to that pull request, it was just part of the same block that had the bit about `DIOCXEND`, but I'll bet this was when it was changed.

miniupnp: I found the change : https://github.com/openbsd/src/commit/072583c99019a91ac511a1e33e10ad02df79ec97
indeed after this commit `pr.nr` is not an input anymore, but an output !

PR #797, opened 2025-01-04
miniupnpd: bind ssdp notify and pcp with v6 to device

qinghon:
Bind ipv6 socket to the interface to completely solve the threat of directional attacks caused by WAN port scanners and leaked IPs

### DIFF
--- a/miniupnpd/minissdp.c
+++ b/miniupnpd/minissdp.c
@@ -402,7 +402,19 @@ OpenAndConfSSDPNotifySocketIPv6(struct lan_addr_s * lan_addr)
 		close(s);
 		return -1;
 	}
-
+#if defined(SO_BINDTODEVICE) && !defined(MULTIPLE_EXTERNAL_IP)
+	/* One and only one LAN interface */
+	if(lan_addrs.lh_first != NULL && lan_addrs.lh_first->list.le_next == NULL
+	   && lan_addrs.lh_first->ifname[0] != '\0')
+	{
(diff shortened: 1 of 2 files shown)


miniupnpd/minissdp.c
```diff
@@ -402,7 +402,19 @@ OpenAndConfSSDPNotifySocketIPv6(struct lan_addr_s * lan_addr)
...
 	}
-
+#if defined(SO_BINDTODEVICE) && !defined(MULTIPLE_EXTERNAL_IP)
+	/* One and only one LAN interface */
+	if(lan_addrs.lh_first != NULL && lan_addrs.lh_first->list.le_next == NULL
+	   && lan_addrs.lh_first->ifname[0] != '\0')
+	{
+		if(setsockopt(s, SOL_SOCKET, SO_BINDTODEVICE,
+		              lan_addrs.lh_first->ifname,
+		              strlen(lan_addrs.lh_first->ifname)) < 0)
```
miniupnp, line 412: shouldn't it be 
```suggestion
		              strlen(lan_addrs.lh_first->ifname) + 1) < 0)
```

miniupnp:
Why Only for IPv6 ?

qinghon:
> Why Only for IPv6 ?

ipv4 is already bound to ip.

miniupnp:
> > Why Only for IPv6 ?
> 
> ipv4 is already bound to ip.

I forgot about that :sweat_smile:

miniupnp (approved):

PR #746, opened 2024-05-28
miniupnpd: Increase port map description length limit to 255

Self-Hosting-Group:
- Same limit as the original Linksys implementation
- Remove the currently inconsistent length limit of 76 for GetGenericPortMappingEntry/GetSpecificPortMappingEntry and 64 for GetListOfPortMappings (IGDv2)
- No longer truncate description to 127 and 89 on daemon restarts, with lease file
- When using pf firewall backend, 63 (pf label) limit remains

The maximum length is not defined in the UPnP IGD standards. The built-in Windows client allows any description length (>20k tested). The PCP description option standard (not implemented) has a length limit of 1016. Other limits: iptables/nftables comments are limited to 256/128 (not used).

PR #812, opened 2025-03-19
miniupnpd: Increase port map description length limit to 255

Self-Hosting-Group:
- Same limit as the original Linksys implementation
- Remove the currently inconsistent length limit of 76 for GetGenericPortMappingEntry/GetSpecificPortMappingEntry and 64 for GetListOfPortMappings (IGDv2)
- No longer truncate description to 127 and 89 on daemon restarts, with lease file
- When using pf firewall backend, 63 (pf label) limit remains

The maximum length is not defined in the UPnP IGD standards. The built-in Windows client allows any description length (>20k tested). The PCP description option standard (not implemented) has a length limit of 1016. Other limits: iptables/nftables comments are limited to 256/128 (not used).

The PR from #746 has been recreated because it raises the limit of the daemon, not the client, and was probably closed by mistake.

### DIFF
--- a/miniupnpd/upnpsoap.c
+++ b/miniupnpd/upnpsoap.c
@@ -790,7 +790,7 @@ GetSpecificPortMappingEntry(struct upnphttp * h, const char * action, const char
 	const char * r_host, * ext_port, * protocol;
 	unsigned short eport, iport;
 	char int_ip[32];
-	char desc[64];
+	char desc[256];
 	unsigned int leaseduration = 0;
 
 	ParseNameValue(h->req_buf + h->req_contentoff, h->req_contentlen, &data);
@@ -1069,7 +1069,7 @@ GetGenericPortMappingEntry(struct upnphttp * h, const char * action, const char 
 	const char * m_index;
 	char * endptr;
 	char protocol[8], iaddr[32];
-	char desc[64];
+	char desc[256];
 	char rhost[40];
 	unsigned int leaseduration = 0;
 	struct NameValueParserData data;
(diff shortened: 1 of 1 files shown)


miniupnp:
the length comes from PF_RULE_LABEL_SIZE which is 64 : https://github.com/openbsd/src/blob/master/sys/net/pfvar.h#L503

however in the linux code (netfilter / netfilter_nft) there is no limit.

xxxxxliil:
@miniupnp 

Reported by Self-Hosting-Group:

This was partially addressed in commit miniupnp/miniupnp@b393e45. The findings:

- When adding a port map with a description containing >=257 characters/bytes, it gets rejected with 501 Action Failed (since some releases; truncated before) with nftables, or additional garbage data (buffer overflow?) appended to the description with iptables, when listing port maps. But after the commit, I would think 1024 characters should work. Also, the description is truncated on a restart, see 320 chars line limit in `reload_from_lease_file`, and miniupnp/miniupnp@20879ab, from 2025-03, and check new CVE's AddPortMapping description vulnerability details in miniupnpd: Core functionality issues https://github.com/Self-Hosting-Group/miniupnpd-issues
- The new reduced body length limit (e.g. with BSD's PF) might be an issue, if I understood the code correctly, as upnpc's minified request is already 341/442 bytes (IGDv1/IGDv2), see `char body[512+MINIUPNPD_DESC_SIZE];` was 1024/2048
- Client side listings of port maps using `upnpc -l/-L` are printed truncated to 70/63 (IGDv1/IGDv2) characters/bytes

As an extra question, I'd like to know what kind of conflict you've had with Self-Hosting-Group? He said there was some unpleasantness between you, which has made him unable to directly raise issues.

Issue #785, opened 2024-10-29
miniupnpd: Current full-cone NAT detection method is not reliable


Headcrabed:
We have seen multiple issues that caused by miniupnpd's current full-cone NAT detection method cannot correctly detect some full-cone case: #772 #705 https://github.com/openwrt/packages/issues/21841 https://github.com/immortalwrt/immortalwrt/discussions/861

And openwrt fw4 maintainer has confirmed that this is not firewall issue:https://github.com/miniupnp/miniupnp/issues/705#issuecomment-2426626751

So I advice that we can change the logic to not just turn off upnp when we cannot detect a full-cone case, but just warn the user in log output and still enable upnp, this would solve the current problem.

davidtall:
same issue ,a CGNAT user

miniupnp:
If there is a bug in the stun code, it should be fixed

Headcrabed:
@miniupnp Sorry to bother you, but is there any progress on this issue?

miniupnp:
> @miniupnp Sorry to bother you, but is there any progress on this issue?

Sorry, I had not been able to find where is the bug in STUN detection implementation.

miniupnp:
related issues #705 / #772

miniupnp:
> @miniupnp Sorry to bother you, but is there any progress on this issue?

@Headcrabed I've spent some time on the issue. miniupnpd does not support a setup where the port is changed.
I've improved the log message so it will be more obvious 114d1bc83197975dc8cdcdc2869392fb1b30d7dd

miniupnp:
with CGNAT, won't simple UPnP proxies be more fit for the job than miniupnpd ?
- https://github.com/phillipberndt/upnprd/
- https://github.com/tomaszmrugalski/portproxy

Headcrabed:
> with CGNAT, won't simple UPnP proxies be more fit for the job than miniupnpd ?
> 
>     * https://github.com/phillipberndt/upnprd/
> 
>     * https://github.com/tomaszmrugalski/portproxy

For this, currently in openwrt, only miniupnp has luci gui to control, so for openwrt users there's basically no choice.
https://openwrt.org/packages/table/start?dataflt%5BName_pkg-dependencies*%7E%5D=upnp

Headcrabed:
> > @miniupnp Sorry to bother you, but is there any progress on this issue?
> 
[...]

Issue #705, opened 2024-01-19
miniupnpd: Unable to make STUN work in a full-cone NAT environment.


daiaji:
```
venv/bin/pystun3 -H stun.qq.com
NAT Type: Full Cone
External IP: 124.228.2.2
External Port: 19485
Press any key to continue
```
```
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: SSDP packet sender [::ffff:192.168.1.11]:34247 (if_index=-1) not from a LAN, ignoring
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: SSDP packet sender [::ffff:192.168.1.11]:53105 (if_index=-1) not from a LAN, ignoring
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: HTTP Connection from 192.168.0.192 closed unexpectedly
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: HTTP Connection from 192.168.0.192 closed unexpectedly
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: HTTP Connection from 192.168.0.192 closed unexpectedly
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: HTTP Connection from 192.168.0.192 closed unexpectedly
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: HTTP Connection from 192.168.0.192 closed unexpectedly
Sat Jan  6 23:30:13 2024 daemon.warn miniupnpd[6990]: HTTP Connection from 192.168.0.192 closed unexpectedly
Sat Jan  6 23:35:16 2024 daemon.notice miniupnpd[6990]: shutting down MiniUPnPd
Sat Jan  6 23:35:16 2024 daemon.notice miniupnpd[7443]: perform_stun: #0 external address or port changed
Sat Jan  6 23:35:16 2024 daemon.notice miniupnpd[7443]: perform_stun: #1 external address or port changed
Sat Jan  6 23:35:16 2024 daemon.notice miniupnpd[7443]: perform_stun: #2 external address or port changed
Sat Jan  6 23:35:16 2024 daemon.notice miniupnpd[7443]: perform_stun: #3 external address or port changed
[...]

PR #813, opened 2025-03-29
miniupnpd: Add ignore_private_ip option

Headcrabed:
In case that router getting private IP but upstream server has full-cone NAT enabled, current full-cone NAT detect mechanism would make miniupnpd stop working. So add force_forwarding option to let user optionally force bypassing this mechanism.

This PR should close https://github.com/miniupnp/miniupnp/issues/785 https://github.com/miniupnp/miniupnp/issues/705

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1025,7 +1025,7 @@ parselanaddr(struct lan_addr_s * lan_addr, const char * str, int debug_flag)
 				INIT_PRINT_ERR("Error parsing address : %s\n", lan_addr->ext_ip_str);
 				return -1;
 			}
-			if(addr_is_reserved(&lan_addr->ext_ip_addr)) {
+			if(addr_is_reserved(&lan_addr->ext_ip_addr) && !GETFLAG(IGNOREPRIVATEIPMASK)) {
 				/* error */
 				INIT_PRINT_ERR("Error: option ext_ip address contains reserved / private address : %s\n", lan_addr->ext_ip_str);
 				return -1;
@@ -1273,6 +1273,10 @@ init(int argc, char * * argv, struct runtime_vars * v)
 #endif
 			case UPNPEXT_IP:
 				use_ext_ip_addr = ary_options[i].value;
+				break;
+			case UPNP_IGNORE_PRIVATE_IP:
+				if(strcmp(ary_options[i].value, "yes") == 0)
+					SETFLAG(IGNOREPRIVATEIPMASK);
 				break;
 			case UPNPEXT_PERFORM_STUN:
 				if(strcmp(ary_options[i].value, "yes") == 0)
@@ -1885,7 +1889,7 @@ init(int argc, char * * argv, struct runtime_vars * v)
 			INIT_PRINT_ERR("Error: option ext_ip contains invalid address %s\n", use_ext_ip_addr);
 			return 1;
 		}
-		if (addr_is_reserved(&addr)) {
+		if (addr_is_reserved(&addr) && !GETFLAG(IGNOREPRIVATEIPMASK)) {
 			INIT_PRINT_ERR("Error: option ext_ip contains reserved / private address %s, not public routable\n", use_ext_ip_addr);
 			return 1;
 		}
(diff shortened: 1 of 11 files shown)


Headcrabed:
> Fixing the STUN detection would IMHO be the better approach. However, this limitation should be removed now that there are certainly more CG-NAT users.

@Self-Hosting-Group I agree with you. Sadly I don’t have an environment for testing detection mechanism now.

@miniupnp This is a long-standing issue, we really need some kind of solutions like this before we can make the detection mechanism working correctly. Also there might be some special use cases that even the fixed detection mechanism won’t work.

Headcrabed:
Besides, I think even if the detection mechanism is fixed, we'd better still keep this option in the code.

Headcrabed:
> Have you considered setting the existing `ext_ip` configuration option as a workaround? This is because the daemon will then know its public IPv4 address and can return it to the clients, which is also a requirement for correct PCP mapping renewals.

In China, there's no static public IP for home users at all, and `ext_ip` would not be a easy-to-use solution in this case since you have to configure it maybe every time when you reboot your router(in some region, ISP would force reset your network link every day, so you have to set `ext_ip` every day). Besides, `force_forwarding` is already being used on x-wrt and immortalwrt and seems has no problem for users. 

But if you want, I can put more information in log to ask users to use stun server or set `ext_ip` when `force_forwarding` is enabled.

Headcrabed:
> > Have you considered setting the existing `ext_ip` configuration option as a workaround? This is because the daemon will then know its public IPv4 address and can return it to the clients, which is also a requirement for correct PCP mapping renewals.
> 
> @Self-Hosting-Group In China, there's no static public IP for home users at all, and `ext_ip` would not be a easy-to-use solution in this case since you have to configure it maybe every time when you reboot your router(in some region, ISP would force reset your network link every day, so you have to set `ext_ip` every day). Besides, `force_forwarding` is already being used on x-wrt and immortalwrt and seems has no problem for users.
> 
> But if you want, I can put more information in log to ask users to use stun server or set `ext_ip` when `force_forwarding` is enabled.

@miniupnp What's your opinion about this?

Headcrabed:
> I discovered two major bugs in the STUN detection code and was able to fix them. As a result, I also implemented a similar but I think better and simpler failback.

@Self-Hosting-Group So you’ll create a new PR to replace mine?

miniupnp:
@Headcrabed sorry for the delay, I want to check a few things in you PR, and maybe improve it.
I'm thinking that maybe "force_forwarding" should better be "ignore_private_ip" ?

Headcrabed:
> @Headcrabed sorry for the delay, I want to check a few things in you PR, and maybe improve it. I'm thinking that maybe "force_forwarding" should better be "ignore_private_ip" ?

@miniupnp seems a nice choice, will update code later.

miniupnpd/testgetifaddr.c
```diff
@@ -18,6 +18,8 @@
 #define LOG_PERROR 0
 #endif
 
+int runtime_flags = 0;
+
```
miniupnp, line 22: it looks useless

miniupnpd/testportinuse.c
```diff
@@ -14,6 +14,8 @@
 #include "config.h"
 #include "portinuse.h"
 
+int runtime_flags = 0;
+
```
miniupnp, line 18: here too...

miniupnpd/miniupnpd.c
```diff
@@ -1025,7 +1025,7 @@ parselanaddr(struct lan_addr_s * lan_addr, const char * str, int debug_flag)
 				INIT_PRINT_ERR("Error parsing address : %s\n", lan_addr->ext_ip_str);
 				return -1;
 			}
-			if(addr_is_reserved(&lan_addr->ext_ip_addr)) {
+			if(addr_is_reserved(&lan_addr->ext_ip_addr) && !GETFLAG(IGNOREPRIVATEIPMASK)) {
```
miniupnp, line 1028: I'm going to print a warning message if the flag is set.

miniupnp (commented):
I'm going t make a few changes

miniupnp:
@Headcrabed I have merged my modified version into master. Just tell me if something is wrong.

Thanks for your contribution !

Headcrabed:
@miniupnp Thanks. And when will miniupnp have a new release? I'm gonna bump upstream openwrt package version to support this PR.

miniupnp:
> @miniupnp Thanks. And when will miniupnp have a new release? I'm gonna bump upstream openwrt package version to support this PR.

https://github.com/miniupnp/miniupnp/releases/tag/miniupnpd_2_3_9

PR #814, opened 2025-04-07
miniunpnpd: Unify naming scheme and rename to ext_ignore_private_ip_check for ignore_private_ip_check

Headcrabed:
Currently we are using a mix of "ignore private ip" and "ignore private ip check" in code. Let's unify them to "ext ignore private ip check". We always use private ip in most cases for internal interfaces, and it would be confusing if we don't use ext prefix.

Besides, full-cone NAT, which is also named as endpoint-independent NAT in some IETF RFC document. Add this name and a source of it to miniupnpd.conf.

### DIFF
--- a/miniupnpd/Changelog.txt
+++ b/miniupnpd/Changelog.txt
@@ -1,5 +1,11 @@
 $Id: Changelog.txt,v 1.529 2025/04/08 21:28:41 nanard Exp $
 
+2025/04/10:
+  Unify naming scheme and rename to ext_allow_private_ipv4 for
+  ignore_private_ip_check
+  Remove the ability to disable port forwarding for
+  update_ext_ip_addr_from_stun function
+
 2025/04/08:
   improve get_wan_connection_status()
 
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1041,7 +1041,7 @@ parselanaddr(struct lan_addr_s * lan_addr, const char * str, int debug_flag)
 				return -1;
 			}
 			if(addr_is_reserved(&lan_addr->ext_ip_addr)) {
-				if (GETFLAG(IGNOREPRIVATEIPMASK)) {
+				if (GETFLAG(ALLOWPRIVATEIPV4MASK)) {
 					syslog(LOG_WARNING, "IGNORED : option ext_ip address contains reserved / private address : %s", lan_addr->ext_ip_str);
 				} else {
 					/* error */
@@ -1176,7 +1176,7 @@ static void update_disable_port_forwarding(void)
 	} else {
 		int reserved = addr_is_reserved(&addr);
 		if (!disable_port_forwarding && reserved) {
-			if (GETFLAG(IGNOREPRIVATEIPMASK)) {
+			if (GETFLAG(ALLOWPRIVATEIPV4MASK)) {
 				syslog(LOG_WARNING, "IGNORED : Reserved / private IP address %s on ext interface %s", if_addr, ext_if_name);
 			} else {
 				syslog(LOG_WARNING, "Reserved / private IP address %s on ext interface %s: Port forwarding is impossible", if_addr, ext_if_name);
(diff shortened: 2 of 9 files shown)


Headcrabed:
@miniupnp Also I noticed you have removed `!GETFLAG(IGNOREPRIVATEIPMASK)` in `upnpredirect.c`, are you sure it's OK?

Headcrabed:
Did some more modifications. I think it's ready now.

miniupnpd/Changelog.txt
```diff
@@ -1,7 +1,7 @@
 $Id: Changelog.txt,v 1.528 2025/04/06 22:30:21 nanard Exp $
 
 2025/04/07:
-  Add a ignore_private_ip_check option in the configuration file to enable
+  Add a ext_ignore_private_ip_check option in the configuration file to enable
```
miniupnp, line 4: shouldn't it be rather `ignore_ext_private_ip_check` ? :sweat_smile:

Headcrabed: @miniupnp Since existing config option all put `ext` at first, I decided to follow this rule. Besides, I'm thinking that `ext_allow_private_ipv4` might be better? This name describes its function more precisely.

Headcrabed:
> @miniupnp Also I noticed you have removed `!GETFLAG(IGNOREPRIVATEIPMASK)` in `upnpredirect.c`, are you sure it's OK?

@miniupnp I looked into the code, seems due to this, we need some extra modifications or this option won't work. I'll push another commit later, please check if it's OK.

miniupnpd/Changelog.txt
miniupnp: great !

miniupnp:
> @miniupnp Also I noticed you have removed `!GETFLAG(IGNOREPRIVATEIPMASK)` in `upnpredirect.c`, are you sure it's OK?

the modification is supposed to prevent `disable_port_forwarding` to be set to 1, right ? so the modification in `upnpredirect.c` shouldn't be there !

Maybe in your case  `disable_port_forwarding` is set to 1 because of the STUN discovery ???

miniupnp:
sorry I have factorized some code in 7fc5b4d2a5e62eb9152441715922b9083c19ed12

Headcrabed:
> > @miniupnp Also I noticed you have removed `!GETFLAG(IGNOREPRIVATEIPMASK)` in `upnpredirect.c`, are you sure it's OK?
> 
> the modification is supposed to prevent `disable_port_forwarding` to be set to 1, right ? so the modification in `upnpredirect.c` shouldn't be there !
> 
> Maybe in your case `disable_port_forwarding` is set to 1 because of the STUN discovery ???

Seems so, is it ok for my third commit?

miniupnp:
@Headcrabed I don't understand why you force `restrictive_nat` to 0. It has nothing to do with having a private ipv4 address, right ?

Headcrabed:
@miniupnp I've updated the code. And https://github.com/miniupnp/miniupnp/commit/7fc5b4d2a5e62eb9152441715922b9083c19ed12 caused test failure. You didn't update `testgetifaddr.sh` together with `testgetifaddr.c`

Headcrabed:
> @Headcrabed I don't understand why you force `restrictive_nat` to 0. It has nothing to do with having a private ipv4 address, right ?

@miniupnp Seems without `restrictive_nat` to 0, `disable_port_forwarding` will still set to 1 when stun server is working correctly+under restrictive NAT, and causing this option won't work. Am I wrong?

miniupnp:
> > @Headcrabed I don't understand why you force `restrictive_nat` to 0. It has nothing to do with having a private ipv4 address, right ?
> 
> @miniupnp Seems without setting `restrictive_nat` to 0, `disable_port_forwarding` will still set to 1 when stun server is working correctly+under restrictive NAT, and causing this option won't work. Am I wrong?

yes but it has nothing to do with having a private IP, restrictive NAT is because it is not a "full cone NAT", right ?

We could either always ignore `restrictive_nat` (use it to only print a warning) or have a specific option to ignore it.

Headcrabed:
> > > @Headcrabed I don't understand why you force `restrictive_nat` to 0. It has nothing to do with having a private ipv4 address, right ?
> > 
> > 
> > @miniupnp Seems without setting `restrictive_nat` to 0, `disable_port_forwarding` will still set to 1 when stun server is working correctly+under restrictive NAT, and causing this option won't work. Am I wrong?
> 
> yes but it has nothing to do with having a private IP, restrictive NAT is because it is not a "full cone NAT", right ?
> 
> We could either always ignore `restrictive_nat` (use it to only print a warning) or have a specific option to ignore it.

Personally I prefer always ignore `restrictive_nat`, since it's not working correctly in some cases such as #785 .

miniupnp:
> Personally I prefer always ignore `restrictive_nat`, since it's not working correctly in some cases such as #785 .

You are right, let just emit some warnings when such a "restrictive NAT" setup is detected.

Headcrabed:
> > Personally I prefer always ignore `restrictive_nat`, since it's not working correctly in some cases such as #785 .
> 
> You are right, let just emit some warnings when such a "restrictive NAT" setup is detected.

@miniupnp you can also close #705 now

PR #819, opened 2025-04-21
NFTables: don't apply iif to rules if USE_IFNAME_IN_RULES is disabled

feral-dot-io:
This small patch skips setting the iif / ifname when using nftables if USE_IFNAME_IN_RULES is not defined. This feature exists for pf and ipf and so I've extended it for my use case with nftables.


miniupnp:
@feral-dot-io thank you !

rebased and merged : c0db681991808adcd9a3e3d698645a780686b2a2 / 24e57b23008a38f5fa8797b0ec6070a1e5222fec

PR #821, opened 2025-04-22
NFTables: add counter to DNAT rules enabling clean_ruleset_interval

feral-dot-io:
Another small patch to add counters to nftable rules for DNAT. This allows `clean_ruleset_interval` to detect rules with no traffic and remove them.

### DIFF
--- a/miniupnpd/Changelog.txt
+++ b/miniupnpd/Changelog.txt
@@ -10,6 +10,9 @@ VERSION 2.3.9 : released on 2025/04/22
     and now setting ext_perform_stun=allow-filtered would allow failed
     STUN CGNAT filter check to be ignored
 
+2025/04/25:
+  NFTables: don't apply iif to rules if USE_IFNAME_IN_RULES is disabled
+
 2025/04/10:
   Unify naming scheme and rename to ext_allow_private_ipv4 for
   ignore_private_ip_check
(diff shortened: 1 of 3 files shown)


feral-dot-io:
I realised I was compiling without the --igd2 option and this removed the need to clean old rules.

This patch does fix an issue but adds a cost of adding a counter to each nftables rule. Is this something you want? Or would you prefer counters to be covered by a #define? Or something else? I'm open to tidying things up if it's helpful.

miniupnpd/Changelog.txt
```diff
@@ -10,6 +10,9 @@ VERSION 2.3.9 : released on 2025/04/22
     and now setting ext_perform_stun=allow-filtered would allow failed
     STUN CGNAT filter check to be ignored
 
+2025/04/25:
+  NFTables: don't apply iif to rules if USE_IFNAME_IN_RULES is disabled
+
```
miniupnp, line 15: ?
I think it is a rebase leftover ;)

miniupnp:
> I realised I was compiling without the --igd2 option and this removed the need to clean old rules.
> 
> This patch does fix an issue but adds a cost of adding a counter to each nftables rule. Is this something you want? Or would you prefer counters to be covered by a #define? Or something else? I'm open to tidying things up if it's helpful.

I don't know the cost of adding that counter ...
https://wiki.nftables.org/wiki-nftables/index.php/Counters

I will add a #ifdef to disable it...

miniupnp:
I cherry picked the commits

feral-dot-io:
Amazing -- thanks.

PR #822, opened 2025-04-28
miniupnpd/pf: fix delete_pinhole for openbsd

mischief:
commit 2be63c9 (pfpinhole.c: use libpfctl, 2024-06-08) added compile time conditions to support libpfctl, but it also mistakenly guarded setting the pfioc_rule action to PF_CHANGE_GET_TICKET.

set the pfioc_rule action once again to PF_CHANGE_GET_TICKET in the non-libpfctl case for OpenBSD. this fixes 'upnpc -D'. without this, we'd get EINVAL from the DIOCCHANGERULE ioctl.


miniupnp:
thanks !

PR #825, opened 2025-05-05
miniupnpd: Fix false negative filtered STUN CGNAT test result

Headcrabed:
This commit has multiple changes:

Fix CGNAT test result for unsupported STUN servers not returning OTHER-ADDRESS / supporting CHANGE-REQUEST's required for endpoint- independent (1:1) CGNAT filtering tests. Since an unsupported server will simply ignore the CHANGE-REQUEST attribute and respond from the same IP/port, it reaches the daemon causing the false negative filtered result before.

For unsupported servers, display an error indicating compatible ones, and then fail.

Simplify STUN CGNAT tests with two different STUN requests. Now, the first (of four) tests server connectivity, while all subsequent (change-IP / change-port set) for an unrestricted endpoint-independent (1:1) CGNAT per RFC 5780 4.4 test I/II, see[1].

Log unknown STUN attributes as debug instead of warning.

STUN IPv4 CGNAT detection briefly described as code comment and existing ones updated.

An important point to note is that the `miniupnpd.conf` example should only suggest STUN servers that are compatible. This is addressed in PR #798, among others.

[1]: https://datatracker.ietf.org/doc/html/rfc5780#section-4.4

### DIFF
--- a/miniupnpd/upnpstun.c
+++ b/miniupnpd/upnpstun.c
@@ -287,6 +287,7 @@ static int parse_stun_response(unsigned char *buffer, size_t len, struct sockadd
 	uint16_t attr_len;
 	int have_address;
 	int have_xor_mapped_address;
+	int have_other_address;
 
 	if (len < 20)
 		return -1;
@@ -310,6 +311,7 @@ static int parse_stun_response(unsigned char *buffer, size_t len, struct sockadd
 	end = buffer + len;
 	have_address = 0;
 	have_xor_mapped_address = 0;
+	have_other_address = 0;
 
 	while (ptr + 4 <= end) {
 
@@ -325,7 +327,7 @@ static int parse_stun_response(unsigned char *buffer, size_t len, struct sockadd
 		switch (attr_type) {
 		case 0x0001:	/* MAPPED-ADDRESS */
 		case 0x0020:	/* XOR-MAPPED-ADDRESS (RFC 5389) */
-		case 0x8020:	/* XOR-MAPPED-ADDRESS (2005 draft) */
+		case 0x8020:	/* Was XOR-MAPPED-ADDRESS (draft 2 of 2005) */
 			/* Mapped Address or XOR Mapped Address */
 			if (attr_len == 8 && ptr[1] == 1) {
 				/* IPv4 address */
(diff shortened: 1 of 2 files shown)


Headcrabed:
This commit is not made by me, and is ported from @Self-Hosting-Group's code.

miniupnpd/upnpstun.c
```diff
@@ -365,32 +367,40 @@ static int parse_stun_response(unsigned char *buffer, size_t len, struct sockadd
 			       attr_len - 4, ptr + 4);
 			}
 			break;
-		case 0x0004:	/* SOURCE-ADDRESS (RFC 3489) */
-		case 0x0005:	/* CHANGED-ADDRESS (RFC 3489) */
 		case 0x802b:	/* RESPONSE-ORIGIN (RFC 5780) */
+		case 0x0004:	/* Was SOURCE-ADDRESS (RFC 3489) */
 		case 0x802c:	/* OTHER-ADDRESS (RFC 5780) */
+		case 0x0005:	/* Was CHANGED-ADDRESS (RFC 3489) */
```
miniupnp, line 373: Why change the order ? :thinking:

miniupnpd/upnpstun.c
```diff
@@ -365,32 +367,40 @@ static int parse_stun_response(unsigned char *buffer, size_t len, struct sockadd
...
 		ptr += attr_len;
 	}
 
-	return have_address ? 0 : -1;
+	if (!have_other_address && have_address) {
+		syslog(LOG_ERR, "ERROR: STUN server not supported, not returning "
+			"OTHER-ADDRESS / support CHANGE-REQUEST's required for "
+			"endpoint-independent (1:1) CGNAT filtering tests per RFC 5780.");
+		syslog(LOG_ERR, "Two public servers: stun.nextcloud.com and stun.sipgate.net");
+		exit(1);
```
miniupnp, line 401: that's radical to exit(1) !

miniupnpd/upnpstun.c
```diff
@@ -430,7 +440,10 @@ int perform_stun(const char *if_name, const char *if_addr, const char *stun_host
 			return -1;
 		}
 
-		fill_request(requests[i], i/2, i%2);
+		/* Determine unrestricted endpoint-independent (1:1) CGNAT in two STUN requests per RFC 5780 4.4 test I/II */
+		/* 1. Connectivity (binding, detect public IPv4), 2. CHANGE-REQUEST with change-IP and change-port set */
+		/* https://datatracker.ietf.org/doc/html/rfc5780#section-4.4 */
+		i == 0 ? fill_request(requests[i], 0, 0) : fill_request(requests[i], 1, 1);
```
miniupnp, line 446: so there are only 2 different requests ?

miniupnp:
> An important point to note is that the miniupnpd.conf example should only suggest STUN servers that are compatible. 

Can you give a list of compatible STUN servers ???

miniupnpd/upnpstun.c
Headcrabed: The idea was not to list the old deprecated STUN attributes first.

miniupnpd/upnpstun.c
Headcrabed: But this is not where we should use exit(0), do you have any better idea?

miniupnpd/upnpstun.c
Headcrabed: Yes, the most direct path to discover an unrestricted endpoint-independent (1:1) CGNAT according to this newer STUN NAT behavior discovery RFC.

Please see the detection procedure in the first three paragraphs of the linked RFC.

Headcrabed:
> > An important point to note is that the miniupnpd.conf example should only suggest STUN servers that are compatible.
> 
> Can you give a list of compatible STUN servers ???

@miniupnp These servers are tested compatible:

```
stun.nextcloud.com
*stun.sipgate.net
stun.3cx.com
*stun.counterpath.com
stun.miwifi.com
stun.nextcloud.com
```
Besides, servers with `*` are not accessible in my country, but passed test when I'm using a vpn.

miniupnpd/upnpstun.c
miniupnp: not exit at all

Headcrabed: @miniupnp Seems removing `exit(1)` would cause the program exit later, so removing it should be ok. But there are some log spamming:

```
miniupnpd[744547]: parse_stun_response: XOR-MAPPED-ADDRESS 1.2.3.4:56455
miniupnpd[744547]: ERROR: STUN server not supported, not returning OTHER-ADDRESS / support CHANGE-REQUEST's required for endpoint-independent (1:1) CGNAT filtering tests per RFC 5780.
miniupnpd[744547]: Two public servers: stun.nextcloud.com and stun.sipgate.net
miniupnpd[744547]: parse_stun_response: Type 0x0101, Length 12, Magic Cookie 2112a442
miniupnpd[744547]: parse_stun_response: XOR-MAPPED-ADDRESS 1.2.3.4:42336
miniupnpd[744547]: ERROR: STUN server not supported, not returning OTHER-ADDRESS / support CHANGE-REQUEST's required for endpoint-independent (1:1) CGNAT filtering tests per RFC 5780.
miniupnpd[744547]: Two public servers: stun.nextcloud.com and stun.sipgate.net
miniupnpd[744547]: parse_stun_response: Type 0x0101, Length 12, Magic Cookie 2112a442
miniupnpd[744547]: parse_stun_response: XOR-MAPPED-ADDRESS 1.2.3.4:39715
miniupnpd[744547]: ERROR: STUN server not supported, not returning OTHER-ADDRESS / support CHANGE-REQUEST's required for endpoint-independent (1:1) CGNAT filtering tests per RFC 5780.
miniupnpd[744547]: Two public servers: stun.nextcloud.com and stun.sipgate.net
miniupnpd[744547]: parse_stun_response: Type 0x0101, Length 12, Magic Cookie 2112a442
miniupnpd[744547]: parse_stun_response: XOR-MAPPED-ADDRESS 1.2.3.4:55942
miniupnpd[744547]: ERROR: STUN server not supported, not returning OTHER-ADDRESS / support CHANGE-REQUEST's required for endpoint-independent (1:1) CGNAT filtering tests per RFC 5780.
miniupnpd[744547]: Two public servers: stun.nextcloud.com and stun.sipgate.net
miniupnpd[744547]: STUN: Performing STUN failed: No such device or address
miniupnpd[744547]: Performing STUN failed. EXITING
```

miniupnpd/upnpstun.c
miniupnp: But then why still have 4 STUN requests if there is only 2 different cases ?

Headcrabed: @miniupnp This is to ensure accurate test results when only a port range is usable with the ISP, continue to test three ports for filtering.

Headcrabed:
@miniupnp any updates?

PR #829, opened 2025-05-14
miniupnpd: fix testportinuse build on bsd with the --portinuse configure option

kurotaka:
testportinuse build fails because proto_itoa() is moved to upnputils.c but the make rule is not updated in Makefile.bsd.
Also update Makefile.macosx and Makefile.sunos since it looks that they require the same fix.


miniupnp (approved):

miniupnp:
@kurotaka thanks !

PR #845, opened 2025-05-31
miniupnpd: leak of snatpmp and snotify

Coeur:
This addresses an Xcode analyser warning regarding the absence of deallocation of snatpmp and snotify on various code paths of miniupnpd.
Capture d’écran 2025-05-31 à 17 51 30

While, sure, at the end of `main`, there will be a free of all that had been allocated by the process, it's still cleaner to:
1. Do the GoodBye, sendto and other shutdown calls.
2. Make the code leak-free for easier re-usability outside of a `main`.

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -2246,6 +2246,7 @@ print_usage:
 int
 main(int argc, char * * argv)
 {
+	int return_code = 0;
 	int i;
 	int shttpl = -1;	/* socket for HTTP */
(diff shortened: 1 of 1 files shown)


miniupnp:
I'm not sure it is good to do the "shutdown" in all startup failure case...

Coeur:
I carefully checked all what was happening in the "shutdown", and it was safe at each stage: the `if` conditions guarding properly if the initializations had happened or not.

PR #847, opened 2025-05-31
miniupnpd: potential dereference of NULL ; potential double free ; missing fclose

Coeur:
This fixes three Xcode analyser warnings.

1. The first one is trivial: a missing fclose on each early return.

The other two are more tricky and depends if `ary_options` is NULL or not.

2. If `ary_options` is not null, then there is a double free: once on `free(ary_options)` then again on `realloc(ary_options, ...)`. Since we want an optimal realloc, we don't want to prematurely add some `ary_options = NULL` right after free. Consequently, it's best to remove the `free`.

3. If `ary_options` is NULL already, then the analyser is confused by a false positive where `num_options` would be non-zero at the start of the loop. To resolve that, we can set `num_options` to zero unconditionally.


miniupnpd/options.c
```diff
@@ -136,11 +136,7 @@ readoptionsfile(const char * fname, int debug_flag)
 	if(!(hfile = fopen(fname, "r")))
 		return -1;
 
-	if(ary_options != NULL)
-	{
-		free(ary_options);
-		num_options = 0;
-	}
+	num_options = 0;
```
miniupnp, line 139: so there is a leak if ary_options is already allocated

Coeur: No leak, it's freed in `freeoptions()`.

PR #848, opened 2025-06-06
Use common exit in ssdpDiscoverDevices function

### LINKED ISSUE #839: Small change in minissdpc.c
While looking at your commit [upnpcommands.h: #include <stddef.h> for size_t](https://github.com/miniupnp/miniupnp/commit/e263ab6f56c382e10fed31347ec68095d691a0e8), tried to find an issue with undefined `size_t`.
There was none, but an old [socklen_t type in MSC](https://github.com/miniupnp/miniupnp/issues/350) topic emerged.

A part of it was not implemented:
https://github.com/miniupnp/miniupnp/blob/e263ab6f56c382e10fed31347ec68095d691a0e8/miniupnpc/src/minissdpc.c#L826-L827
These lines could be replaced with `goto error;`, because `devlist` is null here.

Even better to rename `error` to something like `close_and_return` as this is the only exit point from the function.

irwir:
Should close #839
Renamed `error` to `close_and_return` because this is common exit from function and matches label naming around line 383.

### DIFF
--- a/miniupnpc/src/minissdpc.c
+++ b/miniupnpc/src/minissdpc.c
@@ -698,7 +698,7 @@ ssdpDiscoverDevices(const char * const deviceTypes[],
 		if(error)
 			*error = MINISSDPC_SOCKET_ERROR;
 		PRINT_SOCKET_ERROR("setsockopt(SO_REUSEADDR,...)");
-		goto error;
+		goto close_and_return;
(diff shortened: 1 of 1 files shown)


miniupnp:
@irwir It looks only cosmetics, or am I missing something ?

PR #853, opened 2025-07-31
miniupnpd: Rewrite permission line parser

yangfl:
Permission line parser rewrites input buffer in-place for parsing, which causes several problems:

* It unnecessarily invalidates input buffer for caller. Actually, you might see the following error message when parsing fails:
  > miniupnpd[1234]: parsing error file /etc/miniupnpd.conf line 16 : allow 1024

  since the hyphen after 1024 is erased by a '\0'.

* It fails to validate token separators. For example, the following line will be accepted: 
```
allow 1024-65535X0.0.0.0/0 1024-65535 all 
                ^ could be any character
```
  and even a potential buffer over-read if the character is '\0', since the parser blindlessly skips it.

* The fifth token is never parsed since it gets a previously written '\0'.

Instead fixing them case-by-case, we rewrite it with DFS in mind, which is much simpler and less error-prone.


miniupnpd/upnppermissions.c
```diff
@@ -91,6 +97,84 @@ unescape_char(const char * s, int * seqlen)
...
+static char *
+get_sep(const char * s)
+{
+	if(!isspace(*s))
+		return NULL;
+	do
+		s++;
+	while(isspace(*s));
+	return (char *) s;
+}
```
miniupnp: why return `char *`  and not ` const char *`  ?

yangfl: It's the same as strstr(): you can pass const or non-const strings and get appropriate results. `const char *result = func(conststr)` and `char *result = func(str)` will work.

PR #871, opened 2025-12-16
fix: improve port forwarding disable logging

### LINKED ISSUE #870: bug: The logging levels for private address warnings are inconsistent
When I tried to use miniupnpd on OpenWrt, I saw a warning about disabling it for private addresses behind CGNAT, but I could only see the relevant handling method by setting the log level to info.

xxxxxliil:
By standardizing warning messages to the 'warning' level, we can provide users with potential solutions without having to set the default log level to 'info'.

fix: miniupnp#870

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1192,16 +1192,17 @@ static void update_disable_port_forwarding(void)
 	} else {
 		int reserved = addr_is_reserved(&addr);
 		if (!disable_port_forwarding && reserved) {
+			syslog(LOG_WARNING, "Reserved / private IP address %s on ext interface %s", if_addr, ext_if_name);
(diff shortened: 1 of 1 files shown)


xxxxxliil:
At the request of @Self-Hosting-Group, the pr is set as a draft.

miniupnp (approved):

xxxxxliil:
Due to abandon @Self-Hosting-Group suggestion and continue using the original pull request due to personal life reasons.

miniupnp:
hum I think we should not add too much to LOG_WARNING.
The user should look for information in the documentation...

xxxxxliil:
> hum I think we should not add too much to LOG_WARNING. The user should look for information in the documentation...

It's recommended to try this only after the documentation has appropriate anchor points, because there will be detailed explanations after a stun failure, but a simple sentence about a failure without stun seems unreasonable.

PR #872, opened 2025-12-16
fix(miniupnpd): remove redundant assignment

xxxxxliil:
The variable `disable_port_forwarding` defaults to 0, and no operations have been observed that would set its value to 1.


Chocobo1:
I believe this PR is actually wrong. See https://github.com/miniupnp/miniupnp/pull/890#issuecomment-4217293862.

miniupnp:
@Chocobo1 I think you're most probably right.

Why have I merged this PR ? :thinking:

PR #879, opened 2026-02-18
miniupnpc: Fix undefined behaviour parsing HTTP response

reddwarf69:
The received buffer wasn't zero terminated, but is then passed to atoi() to retrieve the HTTP status code, triggering undefined behaviour.

Detected by Address Sanitizer with strict_string_checks.


miniupnp:
I should probably get rid of atoi() and switch to strtol()

PR #890, opened 2026-04-09
Re-enable port forwarding properly

Chocobo1:
Before this patch the port forwarding cannot be re-enabled once it is disabled.
Revert a0ee71e9fa66b60052bb3d2cf84380b79db3f8c8.

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1199,6 +1199,7 @@ static void update_disable_port_forwarding(void)
 			}
 		} else if (disable_port_forwarding && !reserved) {
 			syslog(LOG_INFO, "Public IP address %s on ext interface %s: Port forwarding is enabled", if_addr, ext_if_name);
+			disable_port_forwarding = 0;
 		}
 	}
 }


Chocobo1:
>Before this patch the port forwarding cannot be re-enabled once it is disabled.

For example, I start up miniupnpd 2.3.10 with `ext_ifname=ppp0` and everything is working. After some time, `ppp0` goes down for a moment and up again later. Now miniupnpd will not be working correctly. It will return `501: Action Failed` for every add port mapping request, which is not expected behavior IMO. See the following log:
```
SOAPAction: urn:schemas-upnp-org:service:WANIPConnection:1#AddPortMapping
AddPortMapping: ext port 44046 to 192.168.1.195:44046 protocol TCP for: app/1.2.3 leaseduration=0 rhost=
UPnP permission rule 0 matched : port mapping accepted
redirecting port 44046 to 192.168.1.195:44046 protocol TCP for: app/1.2.3
Returning UPnPError 501: Action Failed
```

After applying this PR and everything is working as expected again.

xxxxxliil:
I'm so sorry, I didn't even notice this issue. Could someone analyze at which step `disable_port_forwarding = 1` is set, and then subsequently set back to 0?
Also, the suggestion to withdraw a commit should begin with ["revert:"](https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits).

miniupnp:
- `disable_port_forwarding` is set to 1 in `update_ext_ip_addr_from_stun()` if a restricted nat is detected.
- it is also set in `update_disable_port_forwarding()` if the network interface is down.

It should be unset if the network interface is up !

miniupnp (approved):

Issue #867, opened 2025-12-02
Socket handling in connecthostport.c


irwir:
Recently merged PR #866 might lead to further changes.

1. Timeouts previously were used on non-Windows platforms. Should the macro be defined also on Windows as the default option and added into Visual Studio projects (and CMake)?
2. `select()` uses `struct timeval timeout` as a parameter, but on Windows it would be `DWORD timeout`. It is possible to define macros so that compilation fails because of undefined structure (USE_GETHOSTBYNAME;MINIUPNPC_SET_SOCKET_TIMEOUT;MINIUPNPC_IGNORE_EINTR).
Hence, the library might need to implement basic checks of correctness and consistency of build configuration.
3. Does it make sense to use assignment of the same values for timeout? Initialization would be sufficient, and in this case `setsockoption` code would need no conditional compilation.

PR #891, opened 2026-05-03
Slim version of timeout handling

irwir:
Initialisation removes multiple assignments of the same values.
Also replaced 'us' for microseconds with correct 'ms' for milliseconds, and there should be space between numbers and units.

### DIFF
--- a/miniupnpc/src/connecthostport.c
+++ b/miniupnpc/src/connecthostport.c
@@ -103,9 +103,10 @@ SOCKET connecthostport(const char * host, unsigned short port,
 #endif /* #ifdef USE_GETHOSTBYNAME */
 #ifdef MINIUPNPC_SET_SOCKET_TIMEOUT
 #ifdef _WIN32
-	DWORD timeout;
+	DWORD timeout = MINIUPNPC_CONNECT_TIMEOUT_IN_MS; /* 3000 ms */
 #else
-	struct timeval timeout;
+	struct timeval timeout = { MINIUPNPC_CONNECT_TIMEOUT_IN_MS / 1000,
+							   (MINIUPNPC_CONNECT_TIMEOUT_IN_MS % 1000) * 1000 }; /* 3 s */
 #endif
 #endif /* #ifdef MINIUPNPC_SET_SOCKET_TIMEOUT */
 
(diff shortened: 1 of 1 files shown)


miniupnpc/src/connecthostport.c
```diff
@@ -103,9 +103,10 @@ SOCKET connecthostport(const char * host, unsigned short port,
 #endif /* #ifdef USE_GETHOSTBYNAME */
 #ifdef MINIUPNPC_SET_SOCKET_TIMEOUT
 #ifdef _WIN32
-	DWORD timeout;
+	DWORD timeout = MINIUPNPC_CONNECT_TIMEOUT_IN_MS; /* 3000 ms */
 #else
-	struct timeval timeout;
+	struct timeval timeout = { MINIUPNPC_CONNECT_TIMEOUT_IN_MS / 1000,
+							   (MINIUPNPC_CONNECT_TIMEOUT_IN_MS % 1000) * 1000 }; /* 3 s */
 #endif
```
miniupnp, line 110: in some case the `struct timeval` is still needed for _WIN32

irwir: > in some case the `struct timeval` is still needed for _WIN32

https://github.com/miniupnp/miniupnp/blob/9de3fdf01fc26101a54324d5477e5bd8dcd98e4f/miniupnpc/src/connecthostport.c#L105-L109

Then the original code might be using undeclared structure.
Also, It was suggested in the issue #867 to add explicit checks to prevent invalid condition combinations that could mix Unix and Windows API calls.

miniupnp: I'm going to check that your code compile under win32.

miniupnp: ah ok, select() is only used if `MINIUPNPC_IGNORE_EINTR` is defined, which is impossible under _WIN32...

my bad

miniupnpc/src/connecthostport.c
```diff
@@ -126,28 +127,11 @@ SOCKET connecthostport(const char * host, unsigned short port,
...
-	/* https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt
-	 * SO_RCVTIMEO DWORD Sets the timeout, in milliseconds, for blocking
-	 * receive calls. */
-	timeout = MINIUPNPC_CONNECT_TIMEOUT_IN_MS; /* milliseconds */
 	if(setsockopt(s, SOL_SOCKET, SO_RCVTIMEO, (const char *)&timeout, sizeof(timeout)) < 0)
-#else
-	timeout.tv_sec = MINIUPNPC_CONNECT_TIMEOUT_IN_MS / 1000;
-	timeout.tv_usec = (MINIUPNPC_CONNECT_TIMEOUT_IN_MS % 1000) * 1000;
-	if(setsockopt(s, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(struct timeval)) < 0)
-#endif
```
miniupnp, line 139: I may have merged too quickly... You let only the WIN32 code here...the linux man says : 

> SO_RCVTIMEO
>     Sets the timeout value that specifies the maximum amount of time an input function waits until it completes. It accepts a timeval structure with the number of seconds and microseconds specifying the limit on how long to wait for an input operation to complete. If a receive operation has blocked for this much time without receiving additional data, it shall return with a partial count or errno set to [EAGAIN] or [EWOULDBLOCK] if no data is received. The default for this option is zero, which indicates that a receive operation shall not time out. This option takes a timeval structure. Note that not all implementations allow this option to be set.

miniupnpc/src/connecthostport.c
```diff
@@ -126,28 +127,11 @@ SOCKET connecthostport(const char * host, unsigned short port,
...
 		PRINT_SOCKET_ERROR("setsockopt SO_RCVTIMEO");
 	}
-#ifdef _WIN32
-	timeout = MINIUPNPC_CONNECT_TIMEOUT_IN_MS; /* milliseconds */
 	if(setsockopt(s, SOL_SOCKET, SO_SNDTIMEO, (const char *)&timeout, sizeof(timeout)) < 0)
-#else
-	timeout.tv_sec = MINIUPNPC_CONNECT_TIMEOUT_IN_MS / 1000;
-	timeout.tv_usec = (MINIUPNPC_CONNECT_TIMEOUT_IN_MS % 1000) * 1000;
-	if(setsockopt(s, SOL_SOCKET, SO_SNDTIMEO, &timeout, sizeof(struct timeval)) < 0)
-#endif
```
miniupnp, line 150: same for SO_SNDTIMEO : 

> SO_SNDTIMEO
>     Sets the timeout value specifying the amount of time that an output function blocks because flow control prevents data from being sent. If a send operation has blocked for this time, it shall return with a partial count or with errno set to [EAGAIN] or [EWOULDBLOCK] if no data is sent. The default for this option is zero, which indicates that a send operation shall not time out. This option stores a timeval structure. Note that not all implementations allow this option to be set.

miniupnpc/src/connecthostport.c
irwir: > I may have merged too quickly... You let only the WIN32 code here...the linux man says :

When `_WIN32` is undefined, then `struct timeval` gets defined, and the line
`if(setsockopt(s, SOL_SOCKET, SO_SNDTIMEO, (const char *)&timeout, sizeof(timeout)) < 0)`
should be still correct - unless compiler dislikes type casts, but this might be tweaked without multiple conditional compilation directives.

PR #895, opened 2026-06-17
miniupnpd: fix CIDR mask parsing in permission rules

pedro0311:
Commit 54ecff0ae445 ("miniupnpd: Rewrite permission line parser") changed the permission parser to parse both IPv4 addresses and masks through get_addr().  For CIDR masks this means values such as /0, /24 and /32 are first passed to inet_aton() as one-part IPv4 addresses and then converted back to a prefix length.

That depends on non-strict inet_aton() behaviour.  On implementations which do not accept one-part IPv4 strings as valid addresses, get_addr() returns NULL and valid rules such as:

    deny 0-65535 0.0.0.0/0 0-65535

are rejected.

Add a dedicated get_mask() helper.  Dotted masks are still parsed with get_addr(), while CIDR prefix lengths are parsed as decimal integers and validated explicitly in the 0..32 range.  This restores /0, /24 and /32 handling without relying on libc-specific inet_aton() extensions.

### DIFF
--- a/miniupnpd/upnppermissions.c
+++ b/miniupnpd/upnppermissions.c
@@ -175,6 +175,30 @@ get_addr(const char * s, struct in_addr * addr, unsigned int * dot_cnt)
 	return s + i;
 }
 
+static const char *
+get_mask(const char * s, struct in_addr * mask)
+{
+	const char * p;
+	u_short n_bits;
+	unsigned int dot_cnt;
+
+	if(!isdigit((unsigned char)*s))
+		return NULL;
+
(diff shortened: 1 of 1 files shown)


pedro0311:
This is needed in example for https://github.com/FreshTomato-Project/freshtomato-arm with toolchain on patched gcc 5.3 and patched uClibc 0.9.33.2.

miniupnp:
I see there is a problem in [get_addr()](https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/upnppermissions.c#L172) `inet_aton` shouldn't be called when `dot_cnt` is < 3.

PR #896, opened 2026-06-17
Fix W4 level warning - a signed initialiser for an unsigned field

irwir:


miniupnpc/src/addr_is_reserved.c
```diff
@@ -30,7 +30,7 @@ typedef unsigned long uint32_t;
 #endif
 
 /* List of IP address blocks which are private / reserved and therefore not suitable for public external IP addresses */
-#define IP(a, b, c, d) (((a) << 24) + ((b) << 16) + ((c) << 8) + (d))
+#define IP(a, b, c, d) ((uint32_t)(((a) << 24) + ((b) << 16) + ((c) << 8) + (d)))
```
miniupnp, line 33: @irwir isn't it better to cast before the shift ?

```suggestion
#define IP(a, b, c, d) (((uint32_t)(a) << 24) + ((uint32_t)(b) << 16) + ((uint32_t)(c) << 8) + (uint32_t)(d))
```

irwir: The code works correctly, we are only suppressing warnings. Hence, the simpler the better.
Unsigned constants in the table would require no type casts at all, but this would be too much fuss in my opinion.

miniupnp (approved):

PR #905, opened 2026-08-15
miniupnpd: re-enable port forwarding with reserved address when ext_allow_private_ipv4 is set

### LINKED ISSUE #904: miniupnpd: port forwarding cannot be re-enabled on CGNAT with ext_allow_private_ipv4=yes after external interface flap
## Environment

- miniupnpd 2.3.10, 2.3.11 and current master (iptables backend, also affects NAT-PMP/PCP)
- CGNAT setup: the external interface gets a reserved address from 100.64.0.0/10, `ext_allow_private_ipv4=yes` is set (as recommended for CGNAT)

## Bug description

When the external interface is briefly down or unconfigured (e.g. PPPoE reconnect / WAN flap), `update_disable_port_forwarding()` sets `disable_port_forwarding = 1`. When the interface comes back with its reserved/CGNAT address, the flag is never cleared, because the re-enabling condition requires a **non-reserved** address:

```c
} else if (disable_port_forwarding && !reserved) {   /* never true on CGNAT */
        ...
        disable_port_forwarding = 0;
}
```

From then on, every `AddPortMapping` (UPnP) and NAT-PMP/PCP mapping request fails with **501 Action Failed**. The failure is completely silent: `upnp_redirect_internal()` returns -1 on `disable_port_forwarding` without logging anything, so syslog shows no error at all. The only way to recover is restarting miniupnpd.

#890 fixed re-enabling for the public-address case (the missing assignment), but CGNAT setups with `ext_allow_private_ipv4=yes` are still stuck forever after any interface flap.

## Steps to reproduce

1. Run miniupnpd behind CGNAT with `ext_allow_private_ipv4=yes`
2. Add a port mapping → works
3. Flap the external interface (`ifdown`/`ifup`, PPPoE reconnect, ...) — it comes back with a (new) reserved CGNAT address
4. Try to add a port mapping → 501 Action Failed, forever, with no error logged

## Expected behavior

Port forwarding should be re-enabled when the external interface gets its address back, since `ext_allow_private_ipv4=yes` explicitly allows operating with a reserved external address.

CrazyHer:
## Description

Fixes #904.

On CGNAT setups running with `ext_allow_private_ipv4=yes`, once `disable_port_forwarding` is set to 1 (which happens whenever the external interface is briefly down or unconfigured, e.g. during a PPPoE reconnect), it is never cleared: the recovery condition requires a non-reserved address, but the CGNAT address is always reserved. All `AddPortMapping` / NAT-PMP / PCP requests then fail forever with 501 Action Failed, without any log message.

This extends the recovery condition so that port forwarding is also re-enabled when the address is reserved and `ext_allow_private_ipv4` is enabled. This complements #890, which restored re-enabling for the public-address case.

## Testing

Tested on an OpenWrt router (aarch64, iptables backend) behind CGNAT (external address in 100.64.0.0/10, `ext_allow_private_ipv4=yes`):

- Before the fix: after an external interface flap, `AddPortMapping` fails with 501 Action Failed forever (silent, no syslog entry), until miniupnpd is restarted
- After the fix: when the interface recovers its reserved address, port forwarding is re-enabled and `AddPortMapping` works again (verified with `upnpc` add/list/delete round-trips against the patched build)

### DIFF
--- a/miniupnpd/miniupnpd.c
+++ b/miniupnpd/miniupnpd.c
@@ -1205,8 +1205,10 @@ static void update_disable_port_forwarding(void)
 				syslog(LOG_INFO, "Public IP address is required by UPnP/PCP/PMP protocols and clients do not work without it");
 				disable_port_forwarding = 1;
 			}
-		} else if (disable_port_forwarding && !reserved) {
-			syslog(LOG_INFO, "Public IP address %s on ext interface %s: Port forwarding is enabled", if_addr, ext_if_name);
+		} else if (disable_port_forwarding &&
+		           (!reserved || GETFLAG(ALLOWPRIVATEIPV4MASK))) {
+			syslog(LOG_INFO, "%s IP address %s on ext interface %s: Port forwarding is enabled",
+			       reserved ? "Reserved / private" : "Public", if_addr, ext_if_name);
 			disable_port_forwarding = 0;
 		}
 	}


CrazyHer:
Thanks for the quick reply! You're right — I only fixed the non-STUN path (`update_disable_port_forwarding()`). The STUN path (`update_ext_ip_addr_from_stun()`) has the same class of problem: as far as I can see, with `ext_perform_stun=allow-filtered` the final `disable_port_forwarding = restrictive_nat;` assignment is skipped entirely, so once the flag is set to 1 during an interface flap, nothing ever clears it.

Since you already have a complete patch prepared, I'll happily defer to your release. If you'd rather merge a community fix instead, I can extend this PR to cover the STUN path as well. Either way, thanks a lot for miniupnpd!

miniupnp (approved):

miniupnp (approved):

